PR #10, opened 2019-11-18
Creating tables, adding request and Google authentication endpoints

mg2239:
### Summary 
This pull request creates the tables needed for the database, adds most of the needed GET and POST endpoints, as well as adds an endpoint to validate an authentication key sent from the client.

### Notes 
Currently not sure how the back end will be deployed, so for now I just created an endpoint using Express.


config.js
```diff
@@ -1,11 +1,7 @@
-const AWS = require('aws-sdk');
 require('dotenv').config();
 
-const config = new AWS.Config({
-  // hardcode the config for now, DO NOT PUSH
```
lsizemore8: Remove this comment, since we already fixed that

tables.js
```diff
@@ -0,0 +1,242 @@
+const AWS = require('aws-sdk');
+const config = require('./config');
+
+AWS.config.update(config);
+
+// function createLocations() {
```
lsizemore8: Understand: Why are these functions commented out and not deleted or uncommented?

dummydata.js
```diff
@@ -0,0 +1,55 @@
+const AWS = require('aws-sdk');
```
lsizemore8: What is the purpose of this file? If it's not needed, delete it. Also, rename it to be more descriptive of the needs

config.js
mg2239: On my end it is removed? I'll try re-committing the file.

PR #21, opened 2020-02-09
Implement Google auth endpoint

mg2239:
### Summary 
This pull request implements the authentication endpoint in accordance to the current API specification.

### Test Plan 
Make a POST request to /auth using a client id, token, email, and specify user type.


server/router/auth.ts
```diff
@@ -18,22 +18,58 @@ const validIDs = [
...
 // Verify an authentication token
-router.post('/', async (req, res) => {
-  const { token, clientID } = req.body;
-  verify(token, clientID)
-    .then(() => res.send({ success: true }))
-    .catch(() => res.send({ success: false }));
+router.post('/', (req, res) => {
+  const {
+    token, clientID, userType, email,
+  } = req.body;
```
chansen424: Why not have this const declaration all on one line?

server/router/auth.ts
```diff
@@ -18,22 +18,58 @@ const validIDs = [
...
+          },
+        };
+        docClient.scan(params, (err, data) => {
+          if (err) {
+            res.send(err);
+          } else {
+            const userList = data.Items;
+            if (typeof userList![0] === 'undefined') {
+              res.send({
+                success: true,
```
chansen424: I'm wondering if 'success' and 'exists' are necessary now. I feel like the information that they convey can be inferred from the existence or non-existence of the 'id'. If we allow 'id' to be null, we can eliminate the need for 'success' and 'exists' because if 'id' is null, then 'success' and 'exists' must be false. This works in the non-error cases.

The only place where 'success' might be necessary is in the error cases. This will allow us to distinguish between the cases that are actually errors and those cases in which someone simply does not have an account yet.

Change the responses sent by this function as you see fit, but aim for uniformity among them.

server/router/auth.ts
mg2239: eslint didn't like it, i can disable that style in the config

server/router/auth.ts
chansen424: The highlighted code section is not exactly what I wanted to show, but I hope that my explanation conveys the point.

mg2239: that makes sense, ill change it

PR #24, opened 2020-02-29
Added dispatcher authentication, added types to reflect schema, created unscheduled ride request

mg2239:
### Summary 
This pull request adds dispatcher authentication to the auth endpoint (and changes the required post body for authentication), types for each table to help enforce data correctness, and modified the POST /active-ride endpoint to create an unscheduled ride with no driver.

### Test Plan 
`npm start` and make POST requests to /auth and /active-ride

### Notes 
Making a request with incorrect data types still seems to put the data into the database, so I will try to figure out how to fix this in a different pull request.

### Breaking Changes  
New authentication body:

```
{
	"token": string,
	"clientID": string,
	"table": string, // "Riders", "Drivers", or "Dispatchers"
	"email": string
}
```


server/router/rider.ts
```diff
@@ -8,6 +8,27 @@ const router = express.Router();
...
+  phoneNumber: string,
+  email: string,
+  accessibilityNeeds: {
+    needsWheelchair: boolean,
+    hasCrutches: boolean,
+    needsAssistant: boolean,
+  },
+  description: string,
+  picture: string,
+  joinDate: number,
```
chansen424: Why is this a number? Should it be a string in some ISO date format?

server/router/rider.ts
```diff
@@ -29,32 +50,31 @@ router.get('/rider/:riderID', (req, res) => {
 // Put a rider in Riders table
 router.post('/riders', (req, res) => {
   const postBody = req.body;
+  const user: Rider = {
```
chansen424: Maybe rename this from user to rider?

server/router/auth.ts
```diff
@@ -47,7 +49,7 @@ router.post('/', (req, res) => {
             res.send(err);
           } else {
             const userList = data.Items;
-            res.send({ id: userList![0].id ?? null });
+            res.send({ id: userList?.length ? userList[0].id : null });
```
chansen424: Is it not possible that an empty list can be provided and you will still access index 0?

server/router/auth.ts
```diff
@@ -16,6 +16,7 @@ const validIDs = [
   '241748771473-e85o2d6heucd28loiq5aacese38ln4l4.apps.googleusercontent.com',
   '346199868830-dfi7n737u4g6ajl3ketot11d1m3n1sr3.apps.googleusercontent.com',
   '322014396101-8u88pc3q00v6dre4doa64psr9349bhum.apps.googleusercontent.com',
+  '407408718192.apps.googleusercontent.com',
```
chansen424: What is this?

server/router/activeride.ts
```diff
@@ -8,11 +8,23 @@ const router = express.Router();
...
 const docClient = new AWS.DynamoDB.DocumentClient();
 
+type ActiveRide = {
+  id: string,
+  startLocation: string,
+  endLocation: string,
+  startTime: number,
+  endTime: number,
+  isScheduled: boolean,
+  riderID: string,
```
chansen424: Should we have an array of riderIDs instead? This will make the management of ride information much more compact and understandable, I think.

server/router/activeride.ts
```diff
@@ -8,11 +8,23 @@ const router = express.Router();
...
+  driverID: string | null,
+  repeatsOn: string[] | null,
```
chansen424: Is this active ride functioning as a ride by itself, or the ride schedule for a given rider? Because if this repeats, the driver is not guaranteed to be the same every time.

server/router/rider.ts
mg2239: i was planning on storing it w/ a unix timestamp but i actually forgot about ISO so i'll change that

server/router/auth.ts
mg2239: when i was testing it before i couldn't access index 0, but for some reason i was able to when trying to add the dispatcher auth, so ill do some more looking into it and see

server/router/auth.ts
mg2239: oops that was a client id i was trying to test authentication with

server/router/activeride.ts
mg2239: yeah, i forgot multiple riders could be on the same ride

server/router/activeride.ts
mg2239: both at the moment, i didn't think much about repeating rides yet as i wanted to get the basic ride functionality done first, but yeah ill probably have to revise the schema sometime to accommodate for it

server/router/activeride.ts
```diff
@@ -8,11 +8,23 @@ const router = express.Router();
...
 
+type ActiveRide = {
+  id: string,
+  startLocation: string,
+  endLocation: string,
+  startTime: string,
+  endTime: string,
+  isScheduled: boolean,
+  riderID: string[],
+  driverID: string[] | null,
```
mg2239: until i think of a better solution, for now the driver IDs will be ordered depending on the days listed in repeatsOn. if repeatsOn doesn't exist, the array will only have one ID

PR #32, opened 2020-04-20
Google Authenticator sign in/out

ljw222:
### Summary 

The purpose of this pull request is to implement Google Authentication, and connect the backend user authentication to the frontend. Authorized users will be brought to the driver's table upon successful login.

### Test Plan 

Tested login cases for both authorized and unauthorized users.

### Breaking Changes  


frontend/src/components/landing.tsx
```diff
@@ -1,17 +1,63 @@
-import React from 'react';
+import React, { useState } from 'react';
 import Header from './header';
 import Footer from './footer';
 import ReadMore from './readmore';
+import { GoogleLogin, GoogleLogout } from 'react-google-login';
```
chansen424, line 5: Whenever you install your dependencies, do not install them globally. Install them locally. Inside the frontend folder, run `npm i -S react-google-login` to add it to the project's dependencies.

frontend/src/components/landing.tsx
```diff
@@ -1,17 +1,63 @@
...
+import { useHistory } from "react-router-dom";
 import '../styles/header.css';
 
+const clientId = "241748771473-0r3v31qcthi2kj09e5qk96mhsm5omrvr.apps.googleusercontent.com";
```
chansen424, line 9: Do not have clientId exposed like this. Put this value in the React environment file that you have and access it from there. Access it in that way in all of the locations that you use it.

frontend/src/components/table.tsx
```diff
@@ -1,6 +1,20 @@
 import React, { useState } from 'react';
+import { GoogleLogout } from 'react-google-login';
 import '../styles/table.css';
+import { useHistory } from "react-router-dom";
 
+const clientId = "241748771473-0r3v31qcthi2kj09e5qk96mhsm5omrvr.apps.googleusercontent.com";
```
chansen424, line 6: Access this from the React environment file.

frontend/public/index.html
```diff
@@ -1,21 +1,22 @@
...
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta name="theme-color" content="#000000" />
-    <meta
-      name="description"
-      content="Web site created using create-react-app"
-    />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
-    <!--
+<script src="https://apis.google.com/js/api.js" async defer></script>
+<meta name="google-signin-client_id" content="241748771473-0r3v31qcthi2kj09e5qk96mhsm5omrvr.apps.googleusercontent.com">
```
chansen424, line 4: Why is this here? We shouldn't expose the clientId to the public. Is this needed here for some reason? If so, we might have to figure out a different way to put this value here.

frontend/public/index.html
```diff
@@ -39,5 +41,8 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-  </body>
-</html>
+</body>
+
+<!-- <script crossorigin src=""></script> -->
```
chansen424, line 46: Remove this comment.

frontend/public/index.html
SamChou19815: If it has to appear on the client side, then there is no way to hide it. I think client id is designed to be public. It's the API key that has to be kept private.

ljw222: removed these lines, wasn't necessary to include them

PR #36, opened 2020-04-26
Rider table page

ljw222:
### Summary 

Implemented rider table display and addition form consistent with rider documentation from backend.

### Test Plan 

Submits new rider request only when all inputs are valid.

### Notes 

If logged in and on rider table page, logging out should bring user back to home page.

### Breaking Changes  


frontend/src/components/rider-table.tsx
```diff
@@ -0,0 +1,348 @@
...
+          allNeeds = allNeeds.concat("Has Crutches, ");
+        } else {
+          allNeeds = allNeeds.concat("Needs Assistant, ");
+        }
+      }
+    });
+    return allNeeds.substr(0, allNeeds.length - 2);
+  }
+
+  function renderTableData(allRiders: Rider[]) {
```
chansen424, line 295: Unsure where they are exactly, but resolve the errors that appear in the browser console. You should be able to log in fine now.

server/package-lock 2.json
```diff
@@ -0,0 +1,2004 @@
+{
```
SamChou19815, line 1: this file shouldn't be here

chansen424: Hmm... Don't know how that got there. Deleted!

PR #40, opened 2020-05-02
Refactored rider endpoints with Dynamoose

mg2239:
### Summary 
This pull request refactors almost all rider endpoints to be written with Dynamoose. This will help make code easier to write and understand. The ride endpoint is unfinished, as the ride schema is currently being reworked.

### Test Plan 
Run the server and make requests to the endpoints. Overall functionality should not be altered.

### DIFF
--- a/server/router/common.ts
+++ b/server/router/common.ts
@@ -1,7 +1,7 @@
 import { Request, Response } from 'express';
 import { DocumentClient } from 'aws-sdk/clients/dynamodb';
 
-export const deleteByID =
+export const deleteByID = (
   (req: Request, res: Response, client: DocumentClient, table: string) => {
     const { id } = req.params;
     const params = {
@@ -15,6 +15,6 @@ export const deleteByID =
         res.send({ id });
       }
     });
-  }
+  });
 
 export default deleteByID;
--- a/server/router/location.ts
+++ b/server/router/location.ts
@@ -1,52 +1,57 @@
 import express from 'express';
 import uuid from 'uuid/v1';
-import AWS from 'aws-sdk';
-import config from '../config';
+import dynamoose from 'dynamoose';
 
 const router = express.Router();
 
-AWS.config.update(config);
-const docClient = new AWS.DynamoDB.DocumentClient();
-
-type Location = {
+type LocationType = {
   id: string,
-  location: string,
+  name: string,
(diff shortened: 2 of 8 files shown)


server/router/location.ts
```diff
@@ -1,52 +1,57 @@
...
+  //   TableName: 'Locations',
+  //   Key: { id },
+  // };
+  // docClient.get(params, (err, data) => {
+  //   if (err) {
+  //     res.send({ err });
+  //   } else {
+  //     res.send(data);
+  //   }
+  // });
```
chansen424, line 34: If anywhere in this PR, there is something that works currently but has not been replaced by dynamoose, leave it as is. Do not remove functionality by commenting things out. Makes the diff much larger than it needs to be right now. Plus some might be in use by the devs working on the apps for testing purposes.

server/router/rider.ts
```diff
@@ -1,120 +1,105 @@
...
     } else {
-      res.send({ data: data.Items });
+      res.send(data);
     }
   });
 });
 
-// Get all upcoming rides for a rider
+// TODO: Get all upcoming rides for a rider
 router.get('/:id/rides', (req, res) => {
```
chansen424, line 84: Could you possibly re-implement this in this PR. Otherwise, leave it as is. Do not remove the functionality in this PR if it will be changed in another.

server/router/common.ts
```diff
@@ -1,7 +1,7 @@
 import { Request, Response } from 'express';
 import { DocumentClient } from 'aws-sdk/clients/dynamodb';
 
-export const deleteByID =
+export const deleteByID = (
```
chansen424, line 4: Will this entire file be needed at all if we are using dynamoose? Maybe consider replacing the delete endpoints in another PR with their dynamoose equivalents?

server/router/location.ts
mg2239: oh thanks for letting me know, i commented this out temporarily but i guess i forgot to put it back

server/router/common.ts
mg2239: in my next PR i'm currently working on adding common functions like getByID, getAll, etc. and rewriting deleteByID using dynamoose, i just thought it would be a lot for this one

PR #42, opened 2020-05-08
Extending common functions and implementing in Rider, Driver, and Dispatcher

mg2239:
### Summary 
This pull request adds Dynamoose model and callback support into the common functions, making database operations much simpler. These functions are used to revise the rider, driver, and dispatcher endpoints, however getting all rides for a user is still not yet refactored.

### Test Plan 
Run the server and make requests to rider, driver, and dispatcher endpoints. Functionality should be the same.

### DIFF
--- a/server/router/common.ts
+++ b/server/router/common.ts
@@ -118,18 +118,3 @@ export const deleteByID = (
       }
     });
   });
-
-export const deleteByIDOld = (
-  (res: Response, client: DocumentClient, id: string, table: string) => {
-    const params = {
-      TableName: table,
-      Key: { id },
-    };
-    client.delete(params, (err, _) => {
-      if (err) {
-        res.send({ err });
-      } else {
-        res.send({ id });
-      }
-    });
-  });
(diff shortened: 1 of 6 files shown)


server/router/common.ts
```diff
@@ -1,9 +1,126 @@
...
+        res.send({ err: { message: `id not found in ${table}` } });
+      } else if (callback) {
+        callback(data);
+      } else {
+        data.delete().then(() => res.send({ id }));
+      }
+    });
+  });
+
+export const deleteByIDOld = (
```
chansen424, line 122: Once you have replaced the last call to deleteByIDOld with the new deleteByID function, simply delete this old function as there is no need for it.

server/router/dispatcher.ts
```diff
@@ -45,6 +45,9 @@ router.post('/', (req, res) => {
 });
 
 // Remove Dispatcher
-router.delete('/:id', (req, res) => deleteByID(req, res, docClient, 'Dispatchers'));
+router.delete('/:id', (req, res) => {
+  const { id } = req.params;
+  deleteByIDOld(res, docClient, id, 'Dispatchers');
```
chansen424, line 50: Why use this old version of the function when you can use the new version?

mg2239: at the time i wasnt sure if the diff would be too big if i added a new model for dispatchers, but i guess it wouldnt be too different

PR #44, opened 2020-05-09
Setup heroku build for server

SamChou19815:
### Summary 

This PR makes the repo buildable on heroku. The diff focuses on backend only, since it's a more urgent need for mobile developers.

Heroku will purge devDependencies once everything is built. Therefore, we cannot rely on the current setup that requires ts-node and typescript. We must compile the server side TS code to JS. To achieve that, I added the `heroku-postbuild` hook that calls `tsc` to emit JS code. We also need to add a Procfile to tell heroku what to run.

### Test Plan 

Find the deployment card, visit the deployment with the `/health-check` endpoint. You should see plain text `OK`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "dependencies": {
         "@types/express": "^4.17.6",
         "@types/jquery": "^3.3.33",
+        "@types/uuid": "^7.0.3",
         "aws-sdk": "^2.656.0",
         "babel-eslint": "^10.1.0",
         "eslint": "^6.8.0",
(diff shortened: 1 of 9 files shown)


server/app.ts
```diff
@@ -28,5 +28,6 @@ app.use('/past-rides', pastride);
 app.use('/vehicles', vehicle);
 app.use('/locations', location);
 app.use('/auth', auth);
+app.get('/health-check', (_, response) => response.status(200).send('OK'));
```
SamChou19815, line 31: This is not just for testing heroku deployment. It's also helpful for debugging in the future.

package.json
```diff
@@ -6,6 +6,7 @@
     "dependencies": {
         "@types/express": "^4.17.6",
         "@types/jquery": "^3.3.33",
+        "@types/uuid": "^7.0.3",
```
SamChou19815, line 9: Have to move this to dependency section to avoid the purging-devDependencies problem of heroku

PR #57, opened 2020-09-06
Move model code to separate directory

mg2239:
### Summary 
This PR moves the model code previously in the router files into a separate model directory. Some code has also been cleaned up.

### Test Plan 

No functionality changes were made, so running the server as usual should produce the same results.


server/package.json
```diff
@@ -20,7 +20,7 @@
         "typescript": "^3.9.7"
     },
     "scripts": {
-        "start": "nodemon --exec \"ts-node\" --watch router app.ts",
+        "start": "nodemon --exec \"ts-node\" --watch router --watch models app.ts",
```
chansen424, line 23: Do you need `--watch` twice?

mg2239: yeah

chansen424 (approved):

PR #62, opened 2020-09-27
Drivers list

mg2239:
### Summary 
This PR implements the list of driver cards for the drivers page. Some parts of the implementation don't match the design, such as the driver's weekly availability, and last name truncation if a name exceeds 16 characters. Additional improvements that will be needed in the future include adding profile pictures and retrieving vehicle information when getting all drivers, instead of having separate requests.

### Test Plan 
Run the frontend and navigate to the drivers page.


jjc387 (approved):
Overall, this PR looks solid! However, I did notice that your Card.tsx is specific to a DriverCard. While there currently isn't another instance where we may need a "Card" visualization, in the future, you may consider abstracting the Card.tsx file even farther so it can be used in any instance and display any type of information (e.g. rider information), rather than information specific to the driver.

PR #63, opened 2020-10-04
Update ride schema, add populate functionality to rides

mg2239:
### Summary 
This PR updates the ride key schema to only use id instead of id and type. This allows the type of the ride to be changed, and the `/rides` endpoint to handle getting all rides and querying the rides. Models are also added to the Ride schema, allowing a ride's rider, driver, and start and end locations to be accessed with one backend call.

### Test Plan 
Run the backend and make requests to endpoints for rides.

### Notes 
Right now queries are being handled by scanning the database, which will likely get slow once more entries are added to the database. Indexes should be added to the database in a future PR to help speed up the queries. Also Dynamoose support for Typescript is not fully implemented, so models in schema have to be casted to `any`.

### Breaking Changes  
The new rides schema changes and endpoint changes are reflected in the Notion API doc.

### DIFF
--- a/server/models/ride.ts
+++ b/server/models/ride.ts
@@ -1,4 +1,7 @@
 import dynamoose from 'dynamoose';
+import { Location, LocationType } from './location';
+import { Rider, RiderType } from './rider';
+import { Driver, DriverType } from './driver';
 
 export enum Type {
   ACTIVE = 'active',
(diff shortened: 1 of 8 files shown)


server/models/ride.ts
```diff
@@ -7,32 +10,31 @@ export enum Type {
...
-  id: {
-    rangeKey: true,
+  type: {
     type: String,
+    enum: Object.values(Type),
   },
-  startLocation: String,
-  endLocation: String,
+  startLocation: Location as any,
+  endLocation: Location as any,
```
chansen424, line 33: So if someone inputted a random address into a text field, we'd have to store it as a Location in the DB? Seems like that could get out of hand. Let me know if I am misunderstanding anything. We'll definitely have to consider this, but it could just be fixed in a future PR.

chansen424 (approved):

PR #65, opened 2020-10-06
Unscheduled Rides Table Component + Rider List

jjc387:
### Summary 

This pull request is the first step towards implementing the "Unscheduled Rides" table, and its components can be applied to restyling the "Rider List". In particular, the panel displaying the passenger's information on the Unscheduled Rides table has been abstracted into its own components so that it can be reused in displaying rider information in the Rider List.

- [x] implemented Unscheduled Rides table
- [x] modified Rider List

### Test Plan 
Run the website and look at the Unscheduled Rides table and Rider List.

Screen Shot 2020-10-06 at 5 45 03 PM
Screen Shot 2020-10-06 at 5 45 23 PM

### Notes 

The "Unscheduled Rides" table currently holds hard-coded data because an endpoint to get the passenger's information has not been implemented yet. 

The styles for tags are hard-coded since the design is still being finalized, but will be updated once design is finalized and dynamic data can be retrieved.

### Breaking Changes  


frontend/src/components/RiderCard/RiderCard.tsx
```diff
@@ -0,0 +1,34 @@
+import './card.css';
```
mg2239, line 1: can you use css modules instead?

frontend/src/components/RiderCard/RiderCard.tsx
```diff
@@ -0,0 +1,34 @@
...
+      const valList = val.split(',');
+      const tag = valList[1];
+      const newVal = valList[0];
+      const tagString = 'tag ';
+      const tagStyle = tagString + tag.toLowerCase();
+      return (<td key={index} className='passInfo cell'>{newVal} <span className={tagStyle}>{tag}</span></td>);
+    }
+    return (<td key={index} className='passInfo cell'>{val}</td>);
+  });
+  return (<>{resultList}</>);
```
mg2239, line 31: this code is kind of hard to understand, is it possible to clean up and/or add comments? for example i think using else if and else in places that just use if would help, and if a button will always be at the end of a card, that could be handled separate from the row values

frontend/src/components/UserTables/UnscheduledTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+  );
+}
+
+const Table = () => {
+  const passengers = [
+    { startTime: '8:20am', endTime: '8:40am', name: 'Rose Lisborn', pickupLocation: 'Eddygate', pickupTag: 'Ctown', dropoffLocation: 'Hollister Hall', dropoffTag: 'West', needs: 'Crutches' },
+    { startTime: '8:30am', endTime: '8:50am', name: 'Rose Lisborn', pickupLocation: 'Eddygate', pickupTag: 'Ctown', dropoffLocation: 'Hollister Hall', dropoffTag: 'West', needs: 'Crutches' },
+    { startTime: '9:10am', endTime: '9:40am', name: 'Rose Lisborn', pickupLocation: 'Eddygate', pickupTag: 'Ctown', dropoffLocation: 'Hollister Hall', dropoffTag: 'West', needs: 'Crutches' },
+    { startTime: '9:30am', endTime: '9:50am', name: 'Rose Lisborn', pickupLocation: 'Eddygate', pickupTag: 'Ctown', dropoffLocation: 'Hollister Hall', dropoffTag: 'West', needs: 'Crutches' },
+    { startTime: '10:10am', endTime: '10:30am', name: 'Rose Lisborn', pickupLocation: 'Eddygate', pickupTag: 'Ctown', dropoffLocation: 'Hollister Hall', dropoffTag: 'West', needs: 'Crutches' },
```
mg2239, line 26: are you creating passengers as a type in the frontend for a ride's rider? if not you can find the rider type in the [backend docs](https://www.notion.so/Carriage-API-Doc-e78892c2e23e42259aa332b5b3136f22). also time in the database is either ISO 8601 formatted or 24 hour representation, i think youll be receiving the ISO 8601 time from the backend in this case so update the code to handle that instead

frontend/src/components/UserTables/UnscheduledTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+      const comma = ',';
+      const pickup = pickupLocation + comma + pickupTag;
+      const dropoff = dropoffLocation + comma + dropoffTag;
+      const inputValues = [name, pickup, dropoff, needs];
+      const tags = [1, 2];
+      return (
+        <tr key={index}>
+          <td className="cell">{timeframe}</td>
+          <td className="cell"><span style={{ fontWeight: 'bold' }}>{startTime}</span> <br></br> <span style={{ color: '#707070' }}>-- {endTime}</span></td>
+          <RiderCard values={inputValues} tags={tags} />
```
mg2239, line 54: since youre using the rider card for the unscheduled rides table, i think giving it a more general name would be better. also i think you could create a specific type for values, something like
```
type value = {
  data: string
  tag?: string
  // other fields if necessary
}
```
i think that would make displaying the table values easier instead of having to check and process multiple things

frontend/src/types/index.ts
```diff
@@ -23,3 +23,30 @@ export type Driver = {
...
+export type Passenger = {
+  startTime: string;
+  endTime: string;
+  name: string;
+  pickupLocation: string;
+  pickupTag: string;
+  dropoffLocation: string;
+  dropoffTag: string;
+  needs: string;
+}
```
mg2239, line 52: update type if needed

mg2239 (changes requested):
there's a few places that need fixing or could be improved, but overall looks good!

frontend/src/components/UserTables/UnscheduledTable.tsx
jjc387: Yup! Since there isn't an endpoint available to get the exact information needed to display, I just hardcoded everything. Once I make the endpoint, I will adjust.

frontend/src/components/RiderCard/RiderCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+  values: Array<RiderTableValue>;
+};
+
+const RiderCard = (props: RiderCardProps) => {
+  const { values } = props;
+
+  const resultList = values.map((val, index) => {
+    const { data, tag, buttonHandler } = val;
+    if (index === 0) { /* first cell */
+      return (<td key={index} className={`${styles.passInfo} ${styles.cell}
```
mg2239, line 16: there should be a package called 'classnames' that lets you chain classnames cleanly:
```
import cn from 'classnames';
<td className={cn(styles.passInfo, styles.cell, styles.firstCell)}>
```
also since these <td> elements are pretty similar to teach other, you could create a component for it and add boolean props to determine if its first, last, or neither
and if you have a really long jsx element being returned, i like to format it like this to make it more readable
```
return (
  <td
    key={index}
    className={cn(styles.passInfo, styles.cell, styles.firstCell)}
  >
    {data}
  </td>
)
```

frontend/src/components/UserTables/RidersTable.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { useState, useEffect } from 'react';
+import RiderCard from '../RiderCard/RiderCard';
 import Form from '../UserForms/RidersForm';
 import { AccessibilityNeeds, Rider } from '../../types';
 import './table.css';
```
mg2239, line 5: use css modules wherever css is used

mg2239 (approved):
good work with the changes! just a couple more things and i think it should be ready to merge

jjc387:
Abstracted farther and created a TableCell component. Renamed to general names.

PR #67, opened 2020-10-08
Remove colons from keys

chansen424:
### Summary 
Remove colons that snuck into addRider function's object keys.


mly32 (approved):
Looks good.

PR #68, opened 2020-10-08
Added location table and form

mly32:
### Summary 

This pull request implements the "add locations" form. As a specific design for where and how admins are to add locations has not been given, the following choices were made:

- A page at `/dashboard/locations` was created with a similar layout to `Riders`.
    - A table displaying current locations (with the option to delete entries).
    - A form to add entries.
- A new type for Locations was defined in `src/types/index.ts`.
- An icon for locations was taken from the `Carriage > Design > Icons-Rider App` google drive folder.

### Test Plan 

1. Head to `/dashboard` and try clicking on the `locations` icon/link in the sidebar.

2. View the location table entries and check with the [backend](https://carriage-web.herokuapp.com/locations).

3. Try adding a location via the form and again check with [backend](https://carriage-web.herokuapp.com/locations).

4. Try deleting the newly added location via the table and again check with [backend](https://carriage-web.herokuapp.com/locations).

image

image

### Notes 

- Added a `hidden2` style in `table.css` to avoid changing the already defined `hidden`. This hides the form-feedback without changing the centering of the form.

- Potentially look over the rider form and table code, as there are some implementation improvements.

- There may be a need to implement alerts (for example when deleting an entry fails). In the places of alerts, `TODO` has been placed as a comment... Should comments like these be removed?

### Breaking Changes  

N/A


frontend/src/components/UserForms/LocationsForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const fieldName = evt.target.name;
+    const fieldValue = evt.target.value;
+    if (fieldName === 'name') {
+      newLocation.name = fieldValue;
+      setValidName(fieldValue.length > 0);
+    } else if (fieldName === 'address') {
+      newLocation.address = fieldValue;
+      setValidAddress(fieldValue.length > 0);
+    }
+    setnewLocation(newLocation);
```
chansen424, line 27: Might be better to create a copy of newLocation and edit the values in the copy. Then at the end, you can set the newLocation to the altered copy.

frontend/src/components/UserForms/LocationsForm.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { useState } from 'react';
+import { Location } from '../../types/index';
+
+type FormProps = {
+  onClick: ((newLocation: Location) => void);
+}
+
+const Form = ({ onClick }: FormProps) => {
+  const [newLocation, setnewLocation] = useState<Location>({
```
chansen424, line 9: setnewLocation -> setNewLocation

frontend/src/components/UserTables/LocationsTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const deleteLocation = (locationId: string) => {
+    const requestOptions = {
+      method: 'DELETE',
+      headers: { 'Content-Type': 'application/json' },
+    };
+    fetch(`/locations/${locationId}`, requestOptions).then((res) => {
+      if (res.status === 200) {
+        setLocations(locations.filter((l) => l.id !== locationId));
+      } else {
+        // TODO: alert
```
chansen424, line 51: Maybe put a console.err here?

frontend/src/components/UserTables/LocationsTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+      body: JSON.stringify({
+        name: newLocation.name,
+        address: newLocation.address,
+      }),
+    };
+    fetch('/locations', requestOptions).then((res) => {
+      if (res.status === 200) {
+        setLocations([...locations, newLocation]);
+      } else {
+        // TODO: alert
```
chansen424, line 37: console.err some message for now.

frontend/src/components/UserTables/LocationsTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setLocations(locationsData.map((location: any) => ({
+      id: location.id,
+      name: location.name,
+      address: location.address,
+    })));
+  };
+
+  useEffect(() => {
+    getExistingLocations();
+  }, []);
```
chansen424, line 22: Nice. Because otherwise, the setLocations call in getExistingLocations would've given an infinite loop.

chansen424 (changes requested):

frontend/src/components/UserTables/LocationsTable.tsx
mly32: I'm not sure if it's only on my end, but I receive an eslint warning `no-console`.

chansen424: If it's only giving a warning, then just ignore it. We can always remove it later.

mly32:
Fixed issues Chris pointed out.

Also noticed that adding new locations did not properly store the new location (id was not set), which lead to problems when deleting newly adding locations (since ids were set to null). This was fixed.

frontend/src/components/UserForms/LocationsForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <div className="addFormDiv">
+        <form className="addForm" onSubmit={handleSubmit}>
+          <div className="formDiv">
+            <label htmlFor="name" className="formLabel">Name: </label >
+            <input
+              type="text"
+              name="name"
+              onChange={handleInput}
+            />
+            <p className={`formFeedback ${validName && 'hidden2'}`}>
```
chansen424, line 48: Wherever you do this in the className, might be worth using the classnames npm package instead. It allows you to conditionally include classes and such. https://www.npmjs.com/package/classnames

chansen424 (changes requested):

frontend/src/components/UserForms/LocationsForm.tsx
mly32: Cool! I didn't know this was a thing.

mly32:
Updated the location form to use classnames.

I noticed classnames was also used in Sidebar.

I used it differently. Here are the differences:
LocationsForm
```
<p className={cn('formFeedback', { hidden2: validAddress })}>
```
Sidebar
```
<div className={
                path === selected
                  ? cn(styles.selected, styles.circle)
                  : styles.circle
              }>
```

I don't think the way it was used in Sidebar is that good...

chansen424 (approved):

PR #70, opened 2020-10-10
Collapsible component

Becky-Hu-2001:
### Summary 

This pull request implements a collapsible component. Specifically, the following features are created:

- A component named Collapsible is created in `frontend/src/components/Collapsible`
- when using the component users have to pass 1) a `title` prop to indicate the theme of this collapsible section and 2) children components to be displayed within this section
- the component will change from collapsed to expanded or the other way around when the user clicks on the title section

### Test Plan 

- Run the website and see 1) if the collapse functionality works 2) if children components are correctly displayed on the website
test


frontend/src/components/Collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [expanded, setExpanded] = useState(false);
+  const icon = expanded ? down : up;
+
+  const changeExpanded = () => {
+    setExpanded(!expanded);
+  }
+
+  return (
+    <div className={styles.collapsible}>
+      <div className={styles.banner} onClick={changeExpanded}>
```
chansen424, line 20: Don't have to do this. But maybe since this function is so simple, you could inline it? Just have onClick={( ) => setExpanded(!expanded)}

frontend/src/components/Collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { up, down } from "./icons";
+
+type CollapsibleSection = {
+  title: string;
+  children: JSX.Element | JSX.Element[];
+}
+
+const Collapsible = ({ title, children }: CollapsibleSection) => {
+  const [expanded, setExpanded] = useState(false);
+  const icon = expanded ? down : up;
```
chansen424, line 12: I think these icons might need to be flipped? To hide the content I think you'd want to click an up button and to reveal content I think you'd want to click a down button. You could double check the designs or ask a designer, whichever works best for you.

chansen424 (changes requested):

chansen424 (approved):

PR #75, opened 2020-10-24
Add a ride modal

mg2239:
### Summary 

This PR implements a modal form to add a new unscheduled ride to the database. Basic form validation is implemented, such as all fields being required before proceeding, ride start and end times must be sequential, and rider's name must be an existing rider. Autocomplete for existing locations is also implemented, with the ability to input a custom location.

Other additions include adding React Hook Form as a dependency and changing the site's font to Roboto.

### Test Plan 
Input data into modal, query rides endpoint to check if ride data has been added successfully.


mly32 (approved):
If possible, split the _Pages.tsx_ into separate files.

The 3rd modal page doesn't exactly line up with the design on Figma:
image
VS
image

I know this doesn't line up directly with the design, but perhaps add a prev button next to the next button so you can go back and forth between slides.

Also, your form lacks feedback (wrong rider name, times aren't sequential, etc.). Maybe just add some feedback text when it isn't true. 

Maybe talk to the designers about some of these issues.

But, it does what it should, so if you feel it's ready, you can merge.

mg2239:
> The 3rd modal page doesn't exactly line up with the design on Figma:
yeah i asked the designers about that but havent received a response yet. i was under the impression that the times were an alternate idea for a pickup and dropoff time input, but since i put them on the first page it didnt think it was needed, and the rider needs are already stored with the rider. ill make the other changes though!

PR #77, opened 2020-10-24
Add organization (CULift and RedRunner) to Rider schema

simranp723:
### Summary 
This PR adds the field "Organization" to the Rider schema, represented as a String enum value {'CULift', 'RedRunner'}. CULift indicates a CULift rider and RedRunner is a faculty rider. 
This PR also enables querying by this Organization field.

### Test Plan 
Run the backend and make requests to rider GET endpoints.

### Breaking Changes  
This database schema change will be reflection in the Notion API doc.

### DIFF
--- a/server/models/rider.ts
+++ b/server/models/rider.ts
@@ -6,6 +6,11 @@ export enum Accessibility {
   WHEELCHAIR = 'Wheelchair',
 }
 
+export enum Organization {
+  RED_RUNNER = 'Red_Runner',
+  CULIFT = 'CULift'
+}
+
 export type RiderType = {
   id: string
   firstName: string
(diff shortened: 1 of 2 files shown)


server/models/rider.ts
```diff
@@ -6,13 +6,19 @@ export enum Accessibility {
...
+}
+
 export type RiderType = {
   id: string
   firstName: string
   lastName: string
   phoneNumber: string
   email: string
   accessibility: Accessibility[]
+  organization: Organization []
```
chansen424, line 21: RedRunner and CULift riders are mutually exclusive. Could you make it so that the field is not an array, enforcing that you  can only set one or the other?

chansen424 (changes requested):

chansen424 (approved):

PR #78, opened 2020-10-25
Added tags to Location Schema

mly32:
### Summary 
-Added tag enum that currently has west, central, north, ctown (college town), and dtown (downtown).
-Updated the locations form and table on the frontend.

### Test Plan 

1. Run the server locally
2. Test adding, viewing, removing on Postman.
3. Replace the proxy with the local server
4. Run the frontend and test adding removing, viewing tags

### Notes 
Location tags are enums which may need to be updated later with more tags.

### Breaking Changes  
Location schema changed

### DIFF
--- a/frontend/src/components/UserTables/LocationsTable.tsx
+++ b/frontend/src/components/UserTables/LocationsTable.tsx
@@ -10,11 +10,13 @@ const Table = () => {
     const locationsData = await fetch('/locations')
       .then((res) => res.json())
       .then((data) => data.data);
-    setLocations(locationsData.map((location: any) => ({
+    const locs = locationsData.map((location: any) => ({
       id: location.id,
       name: location.name,
       address: location.address,
-    })));
+      ...(location.tag && { tag: location.tag }),
+    }));
+    setLocations(locs);
   };
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


server/models/location.ts
```diff
@@ -1,15 +1,28 @@
 import dynamoose from 'dynamoose';
 
+export enum Tag {
+  WEST = 'west',
+  CENTRAL = 'central',
+  NORTH = 'north',
+  CTOWN = 'ctown', // college town
+  DTOWN = 'dtown', // downtown
+}
```
chansen424, line 9: You already defined this enum in `/types/index.ts`

frontend/src/components/UserTables/LocationsTable.tsx
```diff
@@ -10,73 +10,84 @@ const Table = () => {
...
+    setLocations(locs);
   };
 
   useEffect(() => {
     getExistingLocations();
   }, []);
 
   const addLocation = (newLocation: Location) => {
+    const body = { ...newLocation };
+    delete body.id;
```
chansen424, line 28: Where is this location getting an `id` that you'd have to delete. This request hasn't even been sent to the backend yet, so there shouldn't be an id. If you need to do this because of types, maybe just have `const {id, body} = {...newLocation}` and just don't use id? Might be a cleaner way, so do what you think is best. Just having a `delete` here doesn't feel right.

frontend/src/components/UserTables/LocationsTable.tsx
```diff
@@ -10,73 +10,84 @@ const Table = () => {
     const locationsData = await fetch('/locations')
       .then((res) => res.json())
       .then((data) => data.data);
-    setLocations(locationsData.map((location: any) => ({
+    const locs = locationsData.map((location: any) => ({
```
chansen424, line 13: Might be worth it just keeping this whole thing wrapped in `setLocations`. No need to save variables if you're just going to use them once.

frontend/src/components/UserTables/LocationsTable.tsx
chansen424: Could also edit the Location type to have `id` as optional but that might cause other issues

chansen424 (changes requested):

frontend/src/components/UserTables/LocationsTable.tsx
mly32: Yah, I originally had it like that, but I was running into some issues with my formatter so I set it like so. I'll move it back though.

frontend/src/components/UserTables/LocationsTable.tsx
mly32: The `id` is there since the function takes in a `Location` from the form.

Nice. I didn't know that was an option. I'll use your method.

server/models/location.ts
mly32: Right, `index.ts` is in the frontend folder while `location.ts` is in the server folder. I'm not sure if there's a way to avoid redefining a type. Is there?

chansen424 (approved):

PR #79, opened 2020-10-25
added Ride fields to enable recurring rides

simranp723:
### Summary 

This PR adds the Rider fields: 
- recurring
- recurringDays
- endDate
- deleted
- edits
in order to enable storing rides that recur once or more per week. This also enables editing and deleting of single instances of a recurring ride without altering the master ride pattern.

### Notes 

These fields are optional right now, so that rides can be scheduled as normal without indicating whether they are recurring. 
The design document about how to create, edit, and schedule recurring vs. non-recurring rides is: https://docs.google.com/document/d/1cgifCNPJsfKUSoQMZOpOtb86dVQPACkhIDtp21PYBXA/edit?usp=sharing

### Breaking Changes  

This database schema change will be reflection in the Notion API doc.

### DIFF
--- a/server/models/ride.ts
+++ b/server/models/ride.ts
@@ -27,6 +27,11 @@ export type RideType = {
   endTime: string,
   rider: RiderType,
   driver?: DriverType,
+  recurring?: boolean,
+  recurringDays?: string[],
+  endDate?: string
+  deleted?: boolean,
+  edits?: string[],
 };
 
 const schema = new dynamoose.Schema({
@@ -48,6 +53,17 @@ const schema = new dynamoose.Schema({
   endTime: String,
   rider: Rider as any,
   driver: Driver as any,
+  recurring: Boolean,
+  recurringDays: {
+    type: Array,
+    schema: [String],
+  },
+  deleted: Boolean,
+  edits: {
+    type: Array,
+    schema: [String],
+  },
+  endDate: String,
 });
 
 export const Ride = dynamoose.model('Rides', schema, { create: false });
--- a/server/models/ride.ts
+++ b/server/models/ride.ts
@@ -28,7 +28,7 @@ export type RideType = {
   rider: RiderType,
   driver?: DriverType,
   recurring?: boolean,
-  recurringDays?: string[],
+  recurringDays?: number[],
   endDate?: string
   deleted?: boolean,
   edits?: string[],
@@ -56,7 +56,7 @@ const schema = new dynamoose.Schema({
   recurring: Boolean,
   recurringDays: {
     type: Array,
-    schema: [String],
+    schema: [Number],
   },
   deleted: Boolean,
   edits: {


server/models/ride.ts
```diff
@@ -27,6 +27,11 @@ export type RideType = {
   endTime: string,
   rider: RiderType,
   driver?: DriverType,
+  recurring?: boolean,
+  recurringDays?: string[],
+  endDate?: string
+  deleted?: boolean,
```
chansen424, line 33: Maybe think of a better name than `deleted` as this can indicate that a ride has been independently edited as well, according to you design document.

chansen424, line 33: Also, could you further explain why you'd need to save _both_ the `deleted: false` and `deleted: true` versions of the ride? Not too clear from your design document why the deleted false case is even needed.

server/models/ride.ts
```diff
@@ -27,6 +27,11 @@ export type RideType = {
...
+  edits?: string[],
```
chansen424, line 34: Could you also explain the reasoning behind wanting to save deleted rides instances to the `edits` array as well?

chansen424 (changes requested):

server/models/ride.ts
simranp723: If a ride is independently edited, 2 new one-time rides are created (1 is deleted indicating the instance that shouldn't be scheduled, and the other is the new modified one that is replacing that instance)

chansen424: Why still store the delete one?

server/models/ride.ts
simranp723: The edits array has all the ride instances that are either being replaced or replacing an instance of the recurring ride. The normal ride instance that is being replaced (deleted = True) has to be stored in the edits so that it is not scheduled

server/models/ride.ts
simranp723: For the one-time rides in a recurring ride's edits, deleted = false is the new instance and deleted = true is the old instance that's being replaced

server/models/ride.ts
```diff
@@ -27,6 +27,11 @@ export type RideType = {
   endTime: string,
   rider: RiderType,
   driver?: DriverType,
+  recurring?: boolean,
+  recurringDays?: string[],
```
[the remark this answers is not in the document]
simranp723: I changed it to number[]

server/models/ride.ts
simranp723: So that we know not to schedule that instance of the recurring ride. So if a ride that is normally Monday at 12 is rescheduled to Tuesday at 12 for 1 week, we have to note the Monday ride that has been rescheduled without deleting the entire recurring ride

simranp723: So this would be needed if a ride is rescheduled before it has been scheduled. If it has already been scheduled then we don't have to store the deleted instance separately

chansen424 (approved):

PR #81, opened 2020-10-27
Rider detail page

Becky-Hu-2001:
### Summary 

This pull request 
- implements the rider detail page based off of the user detail component. It can be accessed through clicking on the table on riders page.
- puts all the icons into one single folder to increase modularity
- moves DriverDetail and RiderDetail under the UserDetail folder

### Test Plan 
Check if the data are successfully passed between pages.


frontend/src/components/RiderDetail/icons/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as phone } from './phone.svg';
+export { default as clock } from './clock.svg';
```
chansen424, line 2: Centralize the icons for the frontend into one icons folder, removing any duplicates. I believe the same icons might be saved in multiple places.

chansen424 (changes requested):
Very clean code. Only thing that needs changing is removing duplicate icons saved to the repo. Hope to flesh these pages out with profile pics and tables next sprint.

chansen424 (approved):
Last set of changes was good. I resolved a couple merge conflicts caused by the branch not being updated, but that was the only issue since I've looked at this last. Good job 👍

PR #82, opened 2020-10-27
Add custom locations to backend

mg2239:
### Summary 
This PR adds the ability to add custom locations for rides to the database. On ride creation, if a location is custom, a new location will be created. When a ride changes to past, the custom locations in the ride will be deleted and replaced with a location named "Custom".

Additional changes include a util file with reusable functions not relating to database operations, and fixing an error with the update locations endpoint.

### Test Plan 
Create a ride with a custom location, check Locations table to make sure location was created, update ride type to past, verify locations have been deleted and replaced with "Custom".

### Breaking Changes  
`startLocation` and `endLocation` can be in two formats:
- No city, state, and zipcode; assumed in Ithaca: 123 Main Street or 123 Main Street, Apartment 2
- Any city, state, and zipcode: 123 Main Street, Ithaca, NY 14850

### DIFF
--- a/server/router/ride.ts
+++ b/server/router/ride.ts
@@ -1,8 +1,10 @@
 import express from 'express';
-import { v4 as uuid } from 'uuid';
+import { v4 as uuid, validate } from 'uuid';
 import { Condition } from 'dynamoose';
 import * as db from './common';
-import { Ride, Status, Type } from '../models/ride';
+import { Ride, RideType, Status, Type } from '../models/ride';
+import { Location } from '../models/location';
+import { formatAddress } from '../util';
 
 const router = express.Router();
 const tableName = 'Rides';
@@ -24,7 +26,7 @@ router.get('/', (req, res) => {
     if (type) {
       condition = condition.where('type').eq(type);
     }
-    if(status) {
+    if (status) {
       condition = condition.where('status').eq(status);
     }
     if (rider) {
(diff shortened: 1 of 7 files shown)


server/util/index.ts
```diff
@@ -0,0 +1,14 @@
+import { parseAddress } from 'addresser';
+
+export function createKeys(property: string, values: string[]) {
+  return values.map((v) => ({ [property]: v }));
+}
+
+export function formatAddress(address: string): string {
+  let addressString = address;
+  if (!address.includes(',')) {
+    addressString += ', Ithaca, NY 14850';
```
simranp723, line 10: This would work in most cases, but what if they enter a custom location that is on campus (zip code is different) or not in Ithaca?

server/router/ride.ts
```diff
@@ -45,31 +47,67 @@ router.get('/', (req, res) => {
...
 router.post('/', (req, res) => {
   const {
-    body: {
-      rider,
-      startLocation,
-      endLocation,
-      startTime,
-      endTime,
-    },
+    body: { rider, startTime, endTime, driver, startLocation, endLocation },
```
simranp723, line 50: nice detail keeping the formatting consistent with the other files

server/util/index.ts
```diff
@@ -0,0 +1,14 @@
+import { parseAddress } from 'addresser';
+
+export function createKeys(property: string, values: string[]) {
+  return values.map((v) => ({ [property]: v }));
+}
+
+export function formatAddress(address: string): string {
+  let addressString = address;
+  if (!address.includes(',')) {
```
simranp723, line 9: Also, what if they enter something like "Apartment Name, Apartment #" that has ',' but not the rest of the address?

simranp723 (approved):

server/util/index.ts
mg2239: i think its expected that all locations on campus are provided for, and this would be just for off campus addresses. ill ask to clarify though

server/util/index.ts
mg2239: good point, ill add a better validation check

PR #84, opened 2020-11-05
completed drag drop and other features

mly32:
__check the latest comment for updated features__

### Summary 

This PR improved the schedule view and made events draggable.

- up/down buttons are disabled (disappear) at the border times of the day. 
- time gutter has a fixed width (view width used to change depending on time)
- moved `CalEvent`, `CalResource`, and temporary data like events and colors to separate file 
- step size and time slots were adjusted to have 5-minute increments
- 3 new features on the schedule view:
  - Events can be dragged around
  - Events can be selected (for now an alert displaying the title is shown)
  - New events can be created
    - Click an open slot to enter a title and a 5-minute event will be created
    - Drag in an open area to control the time range and enter a title to create an event longer than 5-minutes

### Test Plan 

Try out each of the newly provided features.

Hidden Button
hidden button

Dragged Event
before
after

Selected Event
selected

New Event
adding
added

### Notes 

- The `any` type was used in some places to avoid problems. I couldn't find a way to do so otherwise.
- I tried playing around with resizeable events but nothing seemed to work

### Breaking Changes  

N/A


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -127,38 +74,69 @@ const Schedule: FunctionComponent = () => {
...
             localizer={localizer}
             events={filterEvents(events)}
             toolbar={false}
-            step={60}
-            timeslots={1}
+            step={5}
+            timeslots={12}
             showMultiDayTimes={true}
             defaultView="day"
+            onEventDrop={onEventDrop}
```
Becky-Hu-2001, line 119: rn if you drag the events, some of them would be resized instead of being dragged to somewhere. Also some other events cannot be resized. I think there's this `onEventResize` that you can implement to address this problem

Becky-Hu-2001 (changes requested):
It's mostly about the DnD's functionalities. Implement `onEventResize` and maybe think about how users can drag the events to somewhere that's not shown on the current calendar view

mly32:
Updated with:

- view more
- drag and drop only allows for a change in the rider (logs the information out)
- no longer supports creating new rides

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -142,6 +136,7 @@ const Schedule = () => {
           </div>
         </div>
       </div>
+      <button className={styles.more} onClick={() => setViewMore(!viewMore)}>view more</button>
```
Becky-Hu-2001, line 139: I think the button text should change to "view less" after the table expands

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -75,39 +79,30 @@ const Schedule = () => {
   };
 
   const onEventDrop = ({ start, end, event, resourceId }: any) => {
+    // eslint-disable-next-line no-console
+    console.log('dragged event:', event.title);
+    // eslint-disable-next-line no-console
+    console.log('old resourceId:', event.resourceId);
+    // eslint-disable-next-line no-console
+    console.log('new resourceId:', resourceId);
```
Becky-Hu-2001, line 87: get rid of the print statement here?

Becky-Hu-2001 (changes requested):
all good, just two minor changes

frontend/src/components/Schedule/Schedule.tsx
mly32: Good catch.

frontend/src/components/Schedule/Schedule.tsx
mly32: Right, this was something Chris requested I add, but it is unnecessary, so I will comment out the console logs and in the future, if anyone needs to view them, they can uncomment. Later down the line, I will remember to delete the logs entirely.

mly32:
Fixed the 2 problems Becky pointed out.

Becky-Hu-2001 (approved):
looks great!

PR #85, opened 2020-11-08
Implement assign driver dropdown

helenxuyang:
### Summary 

This pull request is the first step towards implementing assigning drivers for unscheduled rides

- [x] added Assign buttons to unscheduled rides table 
- [x] implemented UI for driver selection dropdown
- [x] connected to driver backend to display all drivers in suggested/available
- [x] implemented clicking off of dropdown to close it

### Test Plan 

Clicked on first row's assign button:
image

Scroll down:
image

Clicked on third row's assign button:
image

### Notes 

- Still need to connect unscheduled rides table to backend before we can actually assign drivers to rides
- Still need to highlight the available drivers in the schedule


frontend/src/components/UserTables/UnscheduledTable.tsx
```diff
@@ -18,17 +18,18 @@ function renderTableHeader() {
 }
 
 const Table = () => {
+  const [openModal, setOpenModal] = useState(-1);
```
chansen424, line 21: Just want to clarify... the number here will correspond to the index in passengers array below?

frontend/src/components/TableComponents/TableCell.tsx
```diff
@@ -10,21 +10,27 @@ type TableCellProps = {
...
+  const { data, index, first, last, tag, buttonHandler, ButtonModal } = props;
   if (first) {
     return (<><td
       key={index}
       className={cn(styles.passInfo, styles.cell, styles.firstCell)}>
       {data}
     </td></>);
   } if (last) {
+    if (buttonHandler && ButtonModal) {
+      return (<><td key={index} className={`${styles.passInfo} ${styles.cell}
```
chansen424, line 26: <> shouldn't be necessary since whole element has a single outer layer, <td>

frontend/src/components/Modal/AssignDriverModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const wrapperRef = useRef(null);
+  useOutsideAlerter(wrapperRef);
+
+  return (
+    <>
+      {isOpen
+        && <div className={styles.modal} ref={wrapperRef}>
+          <h1 className={styles.titleText}>Suggested Drivers</h1>
+          {drivers.map((driver) => <DriverRow firstName={driver.firstName} imageURL='https://www.biography.com/.image/t_share/MTE5NDg0MDYwNjkzMjY3OTgz/terry-crews-headshot-600x600jpg.jpg' />)}
```
chansen424, line 62: You can get rid of the Suggested section for now since we don't know what suggested would even mean. We might add it back in v2. This has been confirmed with the designers.

frontend/src/components/Modal/assigndrivermodal.module.css
```diff
@@ -0,0 +1,43 @@
...
+
+.driverImage {
+  margin: 7px 14px 6px 27px;
+  border-radius: 50%;
+  width: 36px;
+  height: 36px;
+}
+
+.modal {
+  position: absolute;
```
chansen424, line 34: You want to make sure that this is always under the button, regardless of the screen size. Absolute positions relative to the closest positioned ancestor (https://www.w3schools.com/css/css_positioning.asp) so maybe try to give the button a position attribute so that the modal is rendered relative to its corresponding button?

chansen424 (changes requested):

frontend/src/components/TableComponents/TableCell.tsx
helenxuyang: Whoops, I just copy pasted from the other rows which had <>. Will remove it from those

frontend/src/components/Modal/AssignDriverModal.tsx
helenxuyang: Got it, will just have the Available Drivers section then

frontend/src/components/UserTables/UnscheduledTable.tsx
helenxuyang: Yep, it basically corresponds to the row number because if I just have a bool then all of the dropdowns for all buttons open at once--is there a better way to do this?

frontend/src/components/Modal/assigndrivermodal.module.css
helenxuyang: That makes a lot more sense now, sorry about that!! Fixed so that it's relative to the button and checked with different window sizes

frontend/src/components/Modal/AssignDriverModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const getExistingDrivers = async () => {
+    const driverData = await fetch('/drivers')
+      .then((res) => res.json())
+      .then((data) => data.data);
+    setDrivers(driverData);
+  };
+
+  useEffect(() => {
+    getExistingDrivers();
+  }, []);
```
chansen424, line 52: Is it possible to get the drivers at a higher level, such as when the table renders instead of when the modal  renders? Because I think it sends a request to get all drivers every time you click an assign button? This would involve lifting up the drivers state and passing it down as props so that the request is only made once and the driver dropdowns just use the same data.

chansen424 (changes requested):

frontend/src/components/Modal/AssignDriverModal.tsx
helenxuyang: We just learned about this in Trends today! Raised it to Home because drivers will be used in both Schedule and Table when they're both connected to backend, and Home is their closest ancestor

frontend/src/components/UserTables/UnscheduledTable.tsx
chansen424: I think this might be the best way. I was just making sure I understood the code.

chansen424 (approved):

chansen424:
We'll make drivers accessible through a context in the future if it saves us requests, cleans up the code, and doesn't interfere with component reuse too much.

PR #86, opened 2020-11-08
upgraded react to 17.0

Becky-Hu-2001:
### Summary 

This pull request updated React to its latest version 17.0 and made corresponding changes to accommodate it.
- updated some packages in package.json
- changed some code to accommodate the updated packages

### Test Plan 
see if this update on React is compatible with all the code we have

### Notes 
If you run `yarn start` you'll run into some errors, which are expected to be addressed react-scripts `4.0.1` (it's not released yet) 
As a way out, remember to go into `node_modules/react-scripts/scripts/utils/verifyTypeScriptSetup.js` and change this chunk of code
 
```
    let result;
    parsedTsConfig = immer(readTsConfig, config => {
      result = ts.parseJsonConfigFileContent(
        config,
        ts.sys,
        path.dirname(paths.appTsConfig)
      );
    });
```

into 

```
    parsedTsConfig = {...readTsConfig};

    const result = ts.parseJsonConfigFileContent(
      parsedTsConfig,
      ts.sys,
      path.dirname(paths.appTsConfig)
    );
```

### Breaking Changes  
This PR made more changes than I expected, mostly because many dependencies haven't kept up with React's update.
Most importantly, React 17.0 is not compatible with Typescript's latest official version but only its `4.1beta` version. When I was testing my code, it can work in the browser, but in vscode there's a lot of red lines. I don't know how to get rid of them but I assume they are caused by the Typescript beta version. Plus, I'm also not sure if we should make this update when there're a lot of unreleased dependencies. Therefore this PR may only serve to provide a preview to how React 17 will look like.


chansen424 (approved):
Had to fix some small things in DriverCards and remove unnecessary changes. Plus we somehow didn't have @types/react-dom installed and upgrading eslint in the base directory was causing issues. I wonder if we should remove some dependencies from the base package.json since both the server and frontend folders already have their own package.jsons.

PR #89, opened 2020-11-10
No-show and late statuses for rides

mg2239:
### Summary 
This pull request adds a no-show and late status options for a ride.

### Test Plan 
Make a PUT request to update a ride, with the status being `no_show` or `late`. Check to see that the ride is updated correctly.


server/models/ride.ts
```diff
@@ -15,6 +15,8 @@ export enum Status {
   ARRIVED = 'arrived',
   PICKED_UP = 'picked_up',
   COMPLETED = 'completed',
+  NO_SHOW = 'no_show',
+  LATE = 'late',
```
chansen424, line 19: Might want to keep track of lateness separately. Because then, once the ride is complete, we will lose the data indicating that the person had been late, which we might want to know later on.

chansen424 (changes requested):

server/models/ride.ts
```diff
@@ -15,12 +15,14 @@ export enum Status {
...
   COMPLETED = 'completed',
+  NO_SHOW = 'no_show',
 }
 
 export type RideType = {
   id: string,
   type: Type,
-  status: Status
+  status: Status,
+  late?: boolean,
```
chansen424, line 25: Is it possible to have this not optional and default to false upon ride creation? Could make things easier when accessing the endpoint from clients because we won't have to do null checks or something like that.

chansen424 (changes requested):

server/models/ride.ts
mg2239: yeah thats true, it should be possible 👍

chansen424 (approved):

PR #93, opened 2020-11-28
Fix driver cards rendering

mg2239:
### Summary 
With the new changes to the Driver model in the backend, rendering the cards results in errors. This PR fixes this issue by replacing deprecated fields `startTime`, `endTime`, and `breaks` with `availability`, as well as removing vehicle fetching as that is handled by the backend now.

### Test Plan 

Run the frontend and navigate to the Drivers tab on the sidebar.

### Breaking Changes  


chansen424 (approved):
Looks good on my end locally, and code looks clean.

PR #94, opened 2020-11-28
connected unscheduled rides to backend

mly32:
### Summary 

Set data from `/rides?type=unscheduled` to be displayed on the unscheduled rides table on the home page.

image

### Test Plan 

View home page.

### Notes 

Fixed an issue with `AssignDriverModal` missing keys.

~~The type for rides is called `RealRide` as there was already a ride type defined, which I'm assuming was used for adding rides.~~
There are many broken rides with invalid data.

### Breaking Changes

The type `Ride` was changed since it was defined but never used elsewhere.

### DIFF
--- a/frontend/src/components/UserTables/UnscheduledTable.tsx
+++ b/frontend/src/components/UserTables/UnscheduledTable.tsx
@@ -18,8 +18,8 @@ function renderTableHeader() {
 }
 
 type TableProps = {
-  drivers: Driver[]
-}
+  drivers: Driver[];
+};
 const Table = ({ drivers }: TableProps) => {
   const [openModal, setOpenModal] = useState(-1);
   const [rides, setRides] = useState<RealRide[]>([]);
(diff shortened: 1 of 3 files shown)


frontend/src/components/UserTables/UnscheduledTable.tsx
```diff
@@ -18,59 +18,84 @@ function renderTableHeader() {
...
+    const y = new Date(b.startTime);
+    if (x < y) return -1;
+    if (x > y) return 1;
+    return 0;
+  };
 
-  function renderTableData(allPassengers: Passenger[]) {
-    let currentTime = '';
+  const getUnscheduledRides = () => {
+    fetch('/rides?type=unscheduled')
```
chansen424, line 36: Clarify with designers whether this should only be showing unscheduled rides for the current day. Because right now we do not show the dates, only the times and if a later date has an earlier time than an earlier date, the table appears to be out of order. If the designers only want the unscheduled rides for "today", the endpoint should be able to do that for you.

frontend/src/types/index.ts
```diff
@@ -75,20 +75,22 @@ export type Location = {
...
-};
-
 export type TableValue = {
   data: string | null;
   tag?: string;
   buttonHandler?: () => void;
   ButtonModal?: () => JSX.Element;
 };
+
+export type RealRide = {
```
chansen424, line 85: Could you think of a better name for this than "RealRide"? Is it possible to just call this Ride, replacing the original type that is already there? Or will that cause errors?

chansen424 (changes requested):

frontend/src/types/index.ts
mly32: I didn't want to cause errors with the existing code using `Ride`, but yes, `RealRide` is not a great name. I'll message the creator of `Ride` type to see if we can get a better name for the two of them.

frontend/src/components/UserTables/UnscheduledTable.tsx
mly32: Good catch. I'm very sure the home page only displays information for "today".

frontend/src/types/index.ts
mly32: From what I can tell, the `Ride` type was defined but was never used. I'm going to switch `RealRide` to `Ride` and if the original `Ride` ever needs to be used, I think it's best that the type is defined locally because adding a ride is only done in one place. If a different approach is needed, the names can always be changed later.

chansen424 (approved):

PR #96, opened 2020-12-01
Add requested dropoff time Ride field

simranp723:
### Summary 
This PR adds a new "requestedEndTime" Ride field that is the rider's requested dropoff time for the ride. This field is different from the existing "endTime" field, which is the actual end time of the ride once scheduled.

Only the requestedEndTime can be input, and the endTime is set as the same value when a new ride is created.

### Test Plan 

Create a new ride and verify that the endTime is set to the requestedEndTime.

### Breaking Changes  

Database schema change (Ride table)


chansen424 (approved):
Nice 👍

PR #97, opened 2020-12-01
Fix css

jjc387:
### Summary 

This pull request makes some CSS fixes

- [x] fixed heading on home page
- [x] table can adjust with screen size

### Test Plan 
Run the application and check the changes work accordingly.

Screen Shot 2020-12-01 at 3 15 07 PM

### Breaking Changes  


chansen424 (changes requested):
image
Button in the header appears to be touching the right side of the screen on my end.

image
Smaller than 1150px wide and the table appears to stop resizing. Could you make this more responsive? Would be ideal if the table could still be displayed on a tablet screen such as an iPad (use the responsive design tool presets in your browser to resize to that dimension)

chansen424 (approved):

PR #98, opened 2020-12-02
backend for Schedule

Becky-Hu-82:
### Summary 

This pull request connected the schedule on home page to the backend

- replaced the placeholder rides and riders with actual rides and riders in backend

### Test Plan 
use postman to add rides and see if they can be well displayed on the schedule
image

### Notes 
- In future make sure the rides of type "active" and "past" have a valid driver value.
- Create rides of at least 30 minutes long to have its full information (location, rider, drive) displayed on screen


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -20,9 +39,43 @@ const Schedule = () => {
...
-  const [events, setEvents] = useState(tempEvents);
+  const [events, setEvents] = useState<CalEvent[]>([]);
+  const [drivers, setDrivers] = useState<Driver[]>([]);
   const [viewState, setviewState] = useState(false);
 
+  useEffect(() => {
+    fetch('/rides')
+      .then((res) => res.json())
+      .then(({ data }) => {
+        console.log(data)
```
mg2239, line 50: remove console

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -20,9 +39,43 @@ const Schedule = () => {
...
+        setEvents(data
+          .filter((ride: any) => ride.type !== "unscheduled" && ride.driver !== undefined)
```
mg2239, line 52: types are located in `components/types/index.ts`, use the ones in there instead of `any`
also a ride can be a past ride, do we still want to display those ones?

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -20,9 +39,43 @@ const Schedule = () => {
...
+          .map((ride: any) => ({
+            id: ride.id,
+            title:
+              "" + ride.startLocation.name + " to " + ride.endLocation.name + "\n" +
+              "Rider: " + ride.rider.firstName + " " + ride.rider.lastName + "\n" +
+              "Driver: " + ride.driver.firstName + " " + ride.driver.lastName,
```
mg2239, line 58: use template literals instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -20,9 +39,43 @@ const Schedule = () => {
...
+            start: new Date(ride.startTime.toString()),
+            end: new Date(ride.endTime.toString()),
+            resourceId: ride.driver.id
+          })),
+        )
+      }
+      );
```
mg2239, line 65: these end brackets/parens can be cleaned up, not sure if you have eslint set up but after formatting this is what it looks like for me
image

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -115,8 +168,8 @@ const Schedule = () => {
             max={
               viewState ? defaultEnd : new Date(curStart.getTime() + 7199999)
             }
-            defaultDate={new Date(2018, 0, 29)} // temp date
-            resources={resourceMap1}
+            defaultDate={new Date(2020, 10, 29)} // temp date
```
mg2239, line 171: why not just use the current date?

mg2239 (approved):
looks good to me! mostly just some formatting changes and it should be good to go

frontend/src/components/Schedule/Schedule.tsx
Becky-Hu-2001: for the type because ride's `driver` can be undefined, setting its type to `Ride`will cause red lines. for the past rides, I'm thinking we want the previous rides in that day still on the calendar (but I might be wrong)

frontend/src/components/Schedule/Schedule.tsx
Becky-Hu-2001: I used this date to do testing (because unlike the current date, there are some rides on this date). I'm thinking about leaving it there just so we can do other testing? but if that's unnecessary I can also set to the current date

mg2239: i think you should use the current date, so the schedule functions as it's supposed to. if you wanna do testing you can easily switch it out locally as needed

frontend/src/components/Schedule/Schedule.tsx
mg2239: yeah that makes sense for past rides
i see the issue, i checked and it doesn't look like your branch has the updated types for `Ride` which should take care of the typing issues. try updating your branch from master and see if that resolves the issue

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -20,9 +40,52 @@ const Schedule = () => {
...
 
+  useEffect(() => {
+    fetch('/rides')
+      .then((res) => res.json())
+      .then(({ data }) => {
+        setEvents(data
+          .filter((ride: Ride) => ride.type !== "unscheduled" && ride.driver !== undefined && ride.startLocation !== undefined)
+          .map((ride: Ride) => ({
+            id: ride.id,
+            title:
```
mg2239, line 56: according to the designs, the title doesn't seem to need the rider and driver names. if you need them for the alert then you could just store them with the event and create the string in onSelectEvent. i think that'll also help simplify onEventDrop.

frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -82,7 +145,18 @@ const Schedule = () => {
     // console.log('dragged event:', event.title);
     // console.log('old resourceId:', event.resourceId);
     // console.log('new resourceId:', resourceId);
-    const nextEvents = events.map((old) => (old.id === event.id ? { ...old, resourceId } : old));
+    const updatedCalDriver = calDrivers.find(d => d.resourceId === resourceId);
+    const oldTitle = events.find(e => (e.id === event.id))?.title;
+    const unchanged = oldTitle?.substr(0, oldTitle?.indexOf("Driver"));
+    const title = `${unchanged}Driver: ${updatedCalDriver?.resourceTitle}`;
```
mg2239, line 151: can update event driver instead of updating title if you make the changes mentioned earlier

mg2239 (approved):

mg2239:
good work with the changes! just a couple minor fixes and optimizing the way events are changed on drop and it should be ready to merge

frontend/src/components/Schedule/Schedule.tsx
Becky-Hu-2001: yeah it makes sense to remove the driver's name since we already know that from the schedule header (which can simplify onEventDrop), but I'm thinking it might be better to just put the riders' names on there because if we want to move a ride under another driver, it might to know which ride is the one we want to move by looking at the riders' names (especially if there's a lot of rides at the same time?)

Becky-Hu-2001: I can go ask the designers about that

PR #99, opened 2020-12-06
Add a driver modal

mg2239:
### Summary 
This PR implements a modal to add a new driver. Format validation for email and phone number is implemented as well. Uploading an image is not implemented, but it is possible to select an image from your computer. 

The design of the modal is slightly different from the Figma file, but the differences were checked with the designers and approved.

### Test Plan 
image
Run the frontend and navigate to the Drivers page. Create a driver, and check the Drivers and Vehicles table to check if the entities were created.


chansen424 (approved):
I just merged in the most updated master and resolve the resulting trivial merge conflicts (didn't affect this PRs content at all). Very clean implementation. Like how you separated everything out and built your contexts.

PR #95, opened 2020-11-29
organized user forms and tables

Becky-Hu-2001:
### Summary 

This pull request implemented a form to add dispatchers and organized other user forms and tables

- [x] implemented a DispatcherTable to display all dispatchers and a DispatcherForm to add new dispatchers
- [x] factored out duplicate code in the other files in UserForms folder and put into TextInput, DropDownInput, and CheckBoxInput
- [x] added restrictions to input fields in the RiderModel to guarantee valid input
- [x] added a new endpoint to get all dispatchers

### Test Plan 
user Postman to test getting, adding, and deleting dispacthers

### Notes 
The new endpoint is not built yet, so right now adding a new dispatcher cannot write into the database. I'll fix that soon.

### Breaking Changes  
removed code duplications from UserForms

<!-- - Database schema change (anything that changes DynamoDB document structure)
<!-- - Other change that could cause problems (Detailed in notes)

PR #104, opened 2020-12-13
Create middleware to add endpoint authentication

mg2239:
### Summary 
This PR adds authorization to all endpoints, so now only authorized users with sufficient permissions can make requests to the API. 

When a user is authenticated by the `/auth` endpoint, the API will now return `{ jwt: string }` containing the JWT for the user to attach to all requests in the Authorization header, in the form `Authorization: Bearer <token>`. This JWT is of the form `{ id: string | null, userType: 'Rider' | 'Driver' | 'Dispatcher' }`, and is signed using a secret key in `.env`.

Endpoints can be restricted to any authenticated user or restricted to a certain `userType` and above by adding middleware of this form to any request: `router.get('/', validateUser('User' | 'Rider' | 'Driver' | 'Dispatcher'), (req, res) => ...)`.

### Test Plan 
Authenticate an account through the `/auth` endpoint, and save the token. Make requests to other endpoints with and without the token, and with an edited token. Only the token given by the API should be valid.

### Notes 
The tokens do not expire, so if they get into the wrong hands there still can be a security issue. I felt that implementing refresh tokens might end up being a sizeable block to frontend development, so I chose to omit them for now. However they should be considered for the future.

### Breaking Changes  
Generally the entire flow for making requests is changed, so frontend will need to implement a way to save tokens on client-side.

### DIFF
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -5,7 +5,7 @@ import jwt from 'jsonwebtoken';
 import { ModelType } from 'dynamoose/dist/General';
 import { Document } from 'dynamoose/dist/Document';
 import { Rider } from '../models/rider';
-import { Dispatcher, DispatcherType } from '../models/dispatcher';
+import { Dispatcher } from '../models/dispatcher';
 import { Driver } from '../models/driver';
 
 const router = express.Router();
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -46,22 +46,27 @@ router.post('/', (req, res) => {
     .then((authRes) => {
       const payload = authRes.getPayload();
       const model = getModel(table);
-      if (payload && payload.aud === clientId && model) {
+      if (payload?.aud === clientId && model) {
         model.scan({ email: { eq: email } }).exec((err, data) => {
           if (err) {
             res.send({ success: false, err: err.message });
-          } else if (data) {
-            // Dynamoose type is incorrect
-            const user: any = data[0];
+          } else if (data?.length) {
+            // Dynamoose incorrectly types data[0] as Document[]
+            type User = { id: string };
+            const { id }: User = data[0] as any;
             const userPayload = {
-              id: user ? user.id : null,
+              id,
               userType: getUserType(table),
             };
             res.send({ jwt: jwt.sign(userPayload, process.env.JWT_SECRET!) });
           } else {
-            res.send({ success: false, err: 'Scan data not found' });
+            res.send({ success: false, err: 'User not found' });
           }
         });
+      } else if (payload?.aud !== clientId) {
+        res.send({ success: false, err: 'Invalid client id' });
+      } else if (!model) {
+        res.send({ success: false, err: 'Table not found' });
       } else {
         res.send({ success: false, err: 'Payload not found' });
       }
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -18,7 +18,6 @@ const validIds = [
   '346199868830-dfi7n737u4g6ajl3ketot11d1m3n1sr3.apps.googleusercontent.com',
   '322014396101-8u88pc3q00v6dre4doa64psr9349bhum.apps.googleusercontent.com',
   '241748771473-0r3v31qcthi2kj09e5qk96mhsm5omrvr.apps.googleusercontent.com',
-  '407408718192.apps.googleusercontent.com', // remove when done
 ];
 
 async function verify(clientId: string, token: string): Promise<LoginTicket> {
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -39,10 +39,7 @@ function getModel(table: string) {
   return tableToModel[table];
 }
 
-function getUserType(table: string, user: any) {
-  if (table === 'Dispatchers') {
-    return (user as DispatcherType).accessLevel;
-  }
+function getUserType(table: string) {
   return table.slice(0, table.length - 1);
 }
 
@@ -62,7 +59,7 @@ router.post('/', (req, res) => {
             const user: any = data[0];
             const userPayload = {
               id: user ? user.id : null,
-              userType: getUserType(table, user),
+              userType: getUserType(table),
             };
             res.send({ jwt: jwt.sign(userPayload, process.env.JWT_SECRET!) });
           } else {
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -39,7 +39,7 @@ function getModel(table: string) {
   return tableToModel[table];
 }
 
-function getRole(table: string, user: any) {
+function getUserType(table: string, user: any) {
   if (table === 'Dispatchers') {
     return (user as DispatcherType).accessLevel;
   }
@@ -62,7 +62,7 @@ router.post('/', (req, res) => {
             const user: any = data[0];
             const userPayload = {
               id: user ? user.id : null,
-              userType: getRole(table, user),
+              userType: getUserType(table, user),
             };
             res.send({ jwt: jwt.sign(userPayload, process.env.JWT_SECRET!) });
           } else {
--- a/server/router/auth.ts
+++ b/server/router/auth.ts
@@ -10,7 +10,7 @@ import { Driver } from '../models/driver';
 
 const router = express.Router();
 
-const validIds = [
+const audience = [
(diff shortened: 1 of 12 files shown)


server/router/auth.ts
```diff
@@ -21,42 +22,53 @@ const validIds = [
...
-            res.send({ err });
+            res.send({ success: false, err: err.message });
+          } else if (data) {
+            // Dynamoose type is incorrect
+            const user: any = data[0];
+            const userPayload = {
+              id: user ? user.id : null,
+              userType: getUserType(table),
+            };
+            res.send({ jwt: jwt.sign(userPayload, process.env.JWT_SECRET!) });
```
chansen424, line 60: Sending like this is good as it will be compatible with both mobile and web 👍

server/util/index.ts
```diff
@@ -14,3 +17,53 @@ export function formatAddress(address: string): string {
...
+}
+
+function isUserValid(userType: UserType, authLevel: UserType) {
+  const priority: { [type in UserType]: number } = {
+    User: 0,
+    Rider: 1,
+    Driver: 1,
+    Dispatcher: 2,
+  };
+  return userType === authLevel || priority[authLevel] < priority[userType];
```
chansen424, line 50: This is just a style thing. But if you wanted to make this line shorter, you could just use `priority[authLevel] <= priority[userType]`

server/util/index.ts
```diff
@@ -14,3 +17,53 @@ export function formatAddress(address: string): string {
...
+  }
+}
+
+function isUserValid(userType: UserType, authLevel: UserType) {
+  const priority: { [type in UserType]: number } = {
+    User: 0,
+    Rider: 1,
+    Driver: 1,
+    Dispatcher: 2,
+  };
```
chansen424, line 49: Could define this outside of the function if you really wanted to. That way there is only ever 1 copy of this Map.

server/router/driver.ts
```diff
@@ -29,23 +33,33 @@ router.get('/:id/profile', (req, res) => {
...
+  db.create(res, driver, (data: DriverType) => {
+    const { locals: { user: { id } } } = res;
+    if (id) {
+      res.send(data);
+    } else {
+      res.send({
+        ...data,
+        jwt: jwt.sign({ id: data.id, userType: 'Driver' }, process.env.JWT_SECRET!),
+      });
+    }
```
chansen424, line 51: This could apply to all of the create endpoints. You mentioned that the jwt id would be null upon creation but that the user could get it later.

This should not matter because a dispatcher will always be the one to create new users of any type. Since this is the case, you know that the person creating the users' accounts are not the users themselves and therefore should never get those users' jwt tokens returned to them.

server/router/driver.ts
```diff
@@ -29,23 +33,33 @@ router.get('/:id/profile', (req, res) => {
 });
 
 // Put a driver in Drivers table
-router.post('/', (req, res) => {
+router.post('/', validateUser('Driver'), (req, res) => {
```
chansen424, line 36: Did you mean for this to be Dispatcher? Don't we only want Dispatchers adding and removing users?

server/router/driver.ts
```diff
@@ -29,23 +33,33 @@ router.get('/:id/profile', (req, res) => {
...
 // Update an existing driver
-router.put('/:id', (req, res) => {
+router.put('/:id', validateUser('Driver'), (req, res) => {
   const { params: { id }, body } = req;
   db.update(res, Driver, { id }, body, tableName);
 });
 
 // Delete an existing driver
-router.delete('/:id', (req, res) => {
+router.delete('/:id', validateUser('Driver'), (req, res) => {
```
chansen424, line 62: Same as mentioned in previous comment. Should this be Dispatcher?

server/router/rider.ts
```diff
@@ -59,24 +62,34 @@ router.get('/:id/favorites', (req, res) => {
 });
 
 // Create a rider in Riders table
-router.post('/', (req, res) => {
+router.post('/', validateUser('Rider'), (req, res) => {
```
chansen424, line 65: Dispatcher?

server/router/rider.ts
```diff
@@ -92,7 +105,7 @@ router.post('/:id/favorites', (req, res) => {
 });
 
 // Delete an existing rider
-router.delete('/:id', (req, res) => {
+router.delete('/:id', validateUser('Rider'), (req, res) => {
```
chansen424, line 108: Dispatcher?

server/router/vehicle.ts
```diff
@@ -2,37 +2,40 @@ import express from 'express';
...
 // Get a vehicle by id
-router.get('/:id', (req, res) => {
+router.get('/:id', validateUser('User'), (req, res) => {
   const { params: { id } } = req;
   db.getById(res, Vehicle, id, tableName);
 });
 
 // Get all vehicles
-router.get('/', (req, res) => db.getAll(res, Vehicle, tableName));
+router.get('/', validateUser('User'), (req, res) => {
```
chansen424, line 17: Don't know if this should be Dispatcher? Don't think there is any reason we should let everyone see the info of every vehicle since they already will have access to info on the specific vehicles they need at any given time.

server/router/rider.ts
```diff
@@ -1,25 +1,28 @@
...
 // Get a rider by id in Riders table
-router.get('/:id', (req, res) => {
+router.get('/:id', validateUser('User'), (req, res) => {
   const { params: { id } } = req;
   db.getById(res, Rider, id, tableName);
 });
 
 // Get all riders
-router.get('/', (req, res) => db.getAll(res, Rider, tableName));
+router.get('/', validateUser('User'), (req, res) => {
```
chansen424, line 20: Dispatcher?

server/router/driver.ts
```diff
@@ -1,22 +1,26 @@
...
 // Get a driver by id in Drivers table
-router.get('/:id', (req, res) => {
+router.get('/:id', validateUser('User'), (req, res) => {
   const { params: { id } } = req;
   db.getById(res, Driver, id, tableName);
 });
 
 // Get all drivers
-router.get('/', (req, res) => db.getAll(res, Driver, tableName));
+router.get('/', validateUser('User'), (req, res) => {
```
chansen424, line 18: Dispatcher?

chansen424 (changes requested):
Just some potential permission changes and some minor optional (potentially) code changes. Definitely let me know if I was misunderstanding the `jwt null` concern you brought up in the PR description!

Great work! 💯

chansen424:
Once #95 gets merged in, there will be an endpoint to get all dispatchers. Make sure you add the middleware to that as well.

mg2239:
> As far as I understand, users are only created via a POST request from the Dispatchers, so I don't think the issue you mentioned will even come up?

Ah I see, I think I just misremembered as I was developing. I'll just add an error response if we get a user that tries to log in that's not in the database.

server/util/index.ts
mg2239: ah i did that to make sure any endpoints for riders can't be accessed by drivers and vice versa, do you think that's necessary?

server/router/auth.ts
```diff
@@ -21,42 +22,53 @@ const validIds = [
...
-        docClient.scan(params, (err, data) => {
+      const model = getModel(table);
+      if (payload && payload.aud === clientId && model) {
+        model.scan({ email: { eq: email } }).exec((err, data) => {
           if (err) {
-            res.send({ err });
+            res.send({ success: false, err: err.message });
+          } else if (data) {
+            // Dynamoose type is incorrect
+            const user: any = data[0];
```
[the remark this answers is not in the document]
mg2239: Dynamoose types `data[0]` as a `Document[]`, even though `data` is a 1D array, so I had to cast it to `any`, but i could just do `const user: User = data[0] as any` instead

server/router/vehicle.ts
mg2239: yeah that's true

server/util/index.ts
chansen424: Oh I see. I'd keep it. It might not be necessary but I think it's a good thing to be able to distinguish between rider and driver permissions.

chansen424 (approved):

PR #107, opened 2020-12-15
S3 profile image upload

simranp723:
### Summary 

This PR enables a Driver's, Rider's, or Dispatcher's profile image to be uploaded to S3 with the object key {Driver/Rider/Dispatcher}/{userId}.
Once the file is uploaded, the user's photoLink field in their DynamoDB table will be automatically set to the link to publicly access and download this image (base64-encoded). 

### Test Plan 

Call /upload on an inputted base64-encoded file buffer with the userId and tableName (Driver, Rider, or Dispatcher).
Use the user's photoLink to display the image.

### Notes 

Bucket versioning is currently disabled so that if a user uploads a new profile image, the photoLink will automatically link to the newest version (existing images will be overwritten).


server/app.ts
```diff
@@ -10,10 +11,12 @@ import ride from './router/ride';
 import vehicle from './router/vehicle';
 import location from './router/location';
 import auth from './router/auth';
+import upload from './router/upload';
 
 const port = process.env.PORT || 3001;
 
 dynamoose.aws.sdk.config.update(config);
+AWS.config.update(config);
```
chansen424, line 19: Is this doing the same thing as the line above it? Looks very similar. If so, remove this line and the import for aws-sdk

server/router/upload.ts
```diff
@@ -0,0 +1,50 @@
...
+import * as AWS from 'aws-sdk';
+import { Condition } from 'dynamoose/dist/Condition';
+import * as db from './common';
+import { Rider } from '../models/rider';
+import { Driver } from '../models/driver';
+import { Dispatcher } from '../models/dispatcher';
+
+const router = express.Router();
+
+const BUCKET_NAME = 'dti-carriage-staging-public';
```
chansen424, line 11: Does it matter if this is publicly here or are there some security measures in place on the AWS side? Should we be putting this in a .env?

chansen424 (changes requested):

server/app.ts
simranp723: oh yeah I removed it

server/router/upload.ts
simranp723: It shouldn't matter since it is secured from the AWS side. In this case I think it was set to have public access - I can't view or edit the specific policy but the dev leads could if we want to make it more secure

server/router/upload.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (s3Err) {
+      res.send(s3Err);
+    } else {
+      res.send(s3data);
+    }
+  });
+
+  const objectLink = `${BUCKET_NAME}.s3.us-east-2.amazonaws.com/${objectKey}`;
+  const operation = { $ADD: { photoLink: [objectLink] } };
+  const condition = new Condition().not().exists('photoLink');
```
mg2239, line 37: does this mean if a user wants to upload a new profile pic, they can't?

mg2239: nevermind, just read the docs

mg2239 (approved):
had to make some fixes because the image wasnt actually getting uploaded to S3, but just the base64 string. should be good now though 👍

PR #112, opened 2021-01-09
Add frontend to Heroku deployment

mg2239:
### Summary 
This PR adds the frontend to the Heroku deployment, making it easier for developers to check if their code works in production, as well as interact with the backend.

In addition, all requests using fetch have been replaced to use axios, due to the ability to include interceptors. An interceptor was added in the AuthManager component to handle transferring the JWT to the backend for each request.

### Test Plan 
Visit the Heroku deployment below and check that all pages are visitable.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,40 +13,128 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.12.5",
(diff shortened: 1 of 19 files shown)


mly32 (approved):
Approving to add auth for frontend

PR #115, opened 2021-02-14
packages moved around, eslint updated, types fixed

mly32:
### Summary 

This pull request is the first step to rewriting the various `package.json` and _eslint_ configs.

- Separated dev and non-dev dependencies in packages
- Moved global _devDependencies_ to the main `package.json`
- Added `npm run type-checker`
- Updated _eslint_ config
- Updated typescript configs

### Test Plan

Run the new npm commands and test the frontend & server.

### Important

There were many lines of code that either didn't conform to the new config or were not correctly typed. In an attempt to get all the issues covered in the frontend, some issues were _eslint-disabled_ while others were fixed, but with questionable design.

It is recommended to redefine some rules for eslint. This, however, is not the purpose of this PR.

On a different note, after running `npm run type-check`, a lot of the backend had errors. As I am not familiar with a lot of the backend code, someone else will need to review it.

I also recommend updating the Github actions CI to also run `npm run type-check` after the backend has been fixed.

### Breaking Changes  

Some types in the frontend were corrected but may break prior assumptions.


mg2239 (commented):
added the typecheck to the CI. if you check the details though there seems to be some issues with importing assets, not sure if thats happening on your end too but its happening on my computer. might be something to do with the tsconfigs. we might also want to disable the no explicit any rule, we're using some libraries that unfortunately don't have flexible or correct types especially on the backend so we need to cast to any sometimes

also the heroku deployment is most likely failing because of the devdependencies, so i think we should just keep them in the dependencies for now.

mg2239:
> there seems to be some issues with importing assets

found the issue in react-app-env.d.ts, you should disable this rule too
image

SamChou19815:
> > there seems to be some issues with importing assets
> 
> found the issue in react-app-env.d.ts, you should disable this rule too
> image

Actually you can use this trick

https://github.com/typescript-eslint/typescript-eslint/issues/600#issuecomment-499979248

mg2239 (approved):

PR #118, opened 2021-02-28
Create rider context

lsizemore8:
### Summary 

This PR creates the rider context. To keep the PRs small, this PR does not actually integrate the context into the web application, it just creates the file. The integration will be done in the next PR, but I wanted to confirm that the structure was correct before integrating. 
- [x] implemented rider context
- [ ] fixed Y

### Test Plan 

Given it was not integrated, it should not have broken anything, but here are screenshots of the homepage, riders section, and settings to show that it did not break

dispatcher_homepage_feb_28
rider_page_feb_28
web_homepage_feb_29

### DIFF
--- a/frontend/src/components/UserTables/RidersContext.tsx
+++ b/frontend/src/components/UserTables/RidersContext.tsx
@@ -3,14 +3,14 @@ import { Rider } from '../../types';
 import { useReq } from '../../context/req';
 
 type ridersState = {
-    getExistingRiders: () => Array<Rider>,
-    refreshRiders: () => Promise<Array<Rider>>
+    getExistingRiders: () => Array<Rider>, 
+    refreshRiders: () => Promise<Array<Rider>> 
 };
 
 const initialState: ridersState = {
-  getExistingRiders: () => [],
-  refreshRiders: async () => [],
-};
+   getExistingRiders: () => [],
+   refreshRiders: async () =>  []
+  };
 const RidersContext = React.createContext(initialState);
 
 export const useRiders = () => React.useContext(RidersContext);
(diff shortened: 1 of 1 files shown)


frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const RidesProvider = ({ children }: RidesProviderProps) => {
+  const [riders, setRiders] = useState<RidersType>([])
+  const getExistingRiders = () => {return riders};
+  const { withDefaults } = useReq();
+  const refreshRiders = async () => {
+    const ridersData:RidersType = await fetch('/api/riders', withDefaults())
+        .then((res) => res.json())
+        .then((data) => data.data);
+        
+      setRiders((prev) => ({ ...prev, ridersData}));
```
mg2239, line 33: should be creating a new array here instead of an object, also should be `[ ...ridersData ]` or else we'll end up with duplicate data, so no need for the callback

frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useState } from 'react';
+import { Rider } from '../../types';
+import { useReq } from '../../context/req';
+
+type ridersState = {
+    getExistingRiders: () => Array<Rider>, 
+    refreshRiders: () => Promise<Array<Rider>>
```
mg2239, line 7: getExistingRiders doesn't necessarily need to be a function, it can just be the `riders` variable that's created from `useState`

also refreshRiders can just have a return type of void, since it'll just update the value of `riders`

frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,56 @@
...
+        .then((res) => res.json())
+        .then((data) => data.data);
+        
+      setRiders((prev) => ({ ...prev, ridersData})); 
+      return riders;
+    };
+    //Initialize the data
+    React.useEffect(()=> {
+      refreshRiders();
+    }, [setRiders, withDefaults]);
```
mg2239, line 39: we only want this to run once, so the dep array should be empty

mg2239 (changes requested):

frontend/src/components/UserTables/RidersContext.tsx
lsizemore8: I think it was always creating a new array since RidersType was just an array but that was definitely not needed at all.

frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+const RidersContext = React.createContext(initialState);
+export const useRiders = () => React.useContext(RidersContext);
+
+type RidesProviderProps = {
+  children: React.ReactNode;
+}
+
+export const RidesProvider = ({ children }: RidesProviderProps) => {
+  const [riders, setRiders] = useState<Array<Rider>>([])
+  const getExistingRiders = riders;
```
mg2239, line 24: just using riders should be fine, having a new variable is a bit redundant

frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const ridersData:Array<Rider> = await fetch('/api/riders', withDefaults())
+        .then((res) => res.json())
+        .then((data) => data.data);
+        
+      setRiders([ ...ridersData]); 
+    };
+    //Initialize the data
+    React.useEffect(()=> {
+      refreshRiders();
+    }, [setRiders]);
```
mg2239, line 36: any reason why setRiders is in the dep array?

mg2239 (changes requested):

frontend/src/components/UserTables/RidersContext.tsx
lsizemore8: Ngl I completely misread the documentation, not sure how I messed that up

frontend/src/components/UserTables/RidersContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const ridersData:Array<Rider> = await fetch('/api/riders', withDefaults())
+        .then((res) => res.json())
+        .then((data) => data.data);
+        
+      setRiders([ ...ridersData]); 
+    };
+    //Initialize the data
+    React.useEffect(()=> {
+      refreshRiders();
+    });
```
mg2239, line 35: lol one last thing, need to add an empty dependency array here, otherwise itll run every time the provider re-renders

mg2239 (approved):
looks good, just make that one last fix

PR #120, opened 2021-03-01
Added response status codes to server

simranp723:
### Summary 

This PR adds status codes to server responses, to differentiate between successful requests and errors from the server or client-side.

### Notes 

Status 200 = OK
Status 400 = Client-side error (ex: invalid request/id not found in table)
Status 500 = Server error


pihucode:
Looks good to me! Will wait for Matthew to confirm.

mg2239 (changes requested):
looks good, but i think there are some requests you missed in auth.ts and util/index.ts, add those in and it should be good

mg2239 (approved):

PR #122, opened 2021-03-01
Update Riders frontend

pihucode:
### Summary
This PR updates the frontend design for the Riders tabs with the designs on Figma.

- [x] Update Sidebar icons with new icons.
- [x] Modify Riders table to match the columns specified in the Figma design.

### Test Plan
Compare frontend site with the designs on Figma.
image


mg2239 (commented):
oh sorry for not being clear, dispatchers are admin so you can use `/api/dispatchers` to get all the admin data

pihucode:
> oh sorry for not being clear, dispatchers are admin so you can use `/api/dispatchers` to get all the admin data

Ah I see! The Admin card designs on Figma have availability and car information but the dispatcher schema doesn't have either so do I leave those out?

mg2239:
> Ah I see! The Admin card designs on Figma have availability and car information but the dispatcher schema doesn't have either so do I leave those out?

good question, im actually checking with nina and bryan about what the user types are and what info is stored. turns out some admin can be drivers, so we might have to wait on this PR for a little

frontend/src/components/TableComponents/TableCell.tsx
```diff
@@ -16,21 +16,37 @@ type TableCellProps = {
...
+    const name = data?.substring(0, lastSpaceIndex); //student's name
+    const splitString = data?.split(' ');
+    const netId = splitString ? ' ' + splitString[splitString?.length-1] : '';
+
     return (<td
       key={index}
       className={cn(styles.passInfo, styles.cell, styles.firstCell)}>
-      {data}
+      <span style={{ fontWeight: 'bold' }}>{name}</span>
+      {netId}
```
mg2239, line 28: the first cell of a table isn't necessarily always going to be a name + netid, i think if you change the type of `data` to be ReactNode you can pass in JSX or strings

mg2239 (approved):
table looks good! i noticed that the sidebar, title of the page, and modal button still say rider, so they should be changed to student to match the designs. 

also if possible, could you fix how the disability column is displayed? the current rider type doesn't look up to date with the current schema, so updating it might break some other things, but you could define a `newRider` type for now, and ill go in after this PR merges to do the full fix.

mg2239 (changes requested):
oops meant to request changes instead

frontend/src/components/TableComponents/TableCell.tsx
```diff
@@ -1,40 +1,15 @@
...
 
 type TableCellProps = {
-  data: string | null;
+  data: string | ReactNode | null;
   index: number;
-  first: boolean;
-  last: boolean;
   tag?: string;
-  buttonHandler?: () => void;
-  ButtonModal?: () => JSX.Element;
```
mg2239, line 13: the button handler is needed for other tables, just not the riders table. also we need the first + last fields for the riders table or else the table cells won't be styled properly

frontend/src/components/TableComponents/TableCell.tsx
```diff
@@ -1,40 +1,15 @@
...
-        <Button onClick={buttonHandler}>{data} {<ButtonModal />}</Button></td>);
-    }
-    if (buttonHandler) {
-      return (<><td key={index} className={`${styles.passInfo} ${styles.cell} 
-      ${styles.lastCell}`}>
-        <Button onClick={buttonHandler}>{data}</Button></td></>);
-    }
-    return (<td key={index} className={`${styles.passInfo} ${styles.cell} 
-    ${styles.lastCell}`}>{data}</td>);
-  } if (tag) {
```
mg2239, line 37: add this code back in

frontend/src/components/TableComponents/TableRow.tsx
```diff
@@ -15,30 +15,14 @@ const TableRow = (props: TableRowProps) => {
...
-      /* last cell */
-      return (
-        <TableCell
-          key={index}
-          data={data}
-          index={index}
-          first={false}
-          last={true}
-          buttonHandler={buttonHandler}
-          ButtonModal={ButtonModal}
```
mg2239, line 41: add this code back in

frontend/src/components/TableComponents/TableRow.tsx
```diff
@@ -47,8 +31,6 @@ const TableRow = (props: TableRowProps) => {
         key={index}
         data={data}
         index={index}
-        first={false}
-        last={false}
```
mg2239, line 51: add this code back

mg2239 (changes requested):

mg2239 (commented):
thanks for the changes! since the table components are shared with every table in the frontend, only delete things if they are not used for any other table. other than that everything looks good to me

mg2239 (approved):

PR #126, opened 2021-03-05
Export scheduled rides into CSV

Becky-Hu-82:
### Summary 

This pull request enables the export of unscheduled rides into csv files by

- using fast-csv to convert today's rides into readable stream,
- using react-csv to download the data 

### Test Plan 

Go to /home/export and click the export button in the bottom, see if the data can correctly output into a csv file.

### Notes 

Right now there is an error message saying `id not found in Rides`, which is something we need to fix.

### Breaking Changes  


mg2239 (changes requested):
code works well, just a few comments:
image
you'll want to process this data more before sending it from the server (i.e remove id, late, status, etc. and make the dates, driver, rider, etc. more readable). remember to look at the spreadsheet i sent for what is needed

also something i didn't notice before is that the export preview page and this request are returning unscheduled rides, but we only want scheduled and past rides

and maybe check with jahin if the download button is okay being at the bottom of the page? there doesn't seem to be a design for where it should be but would be good to double check

frontend/src/components/ExportPreview/ExportPreview.tsx
```diff
@@ -30,6 +47,16 @@ const ExportPreview = () => {
           />
         ))}
       </div>
+      <div className={styles.exportButtonContainer}>
+        <ExportButton onClick={downloadCSV}/>
+        <CSVLink
+         data={downloadData}
+         filename='scheduledRides.csv'
```
mg2239, line 54: include the date in the filename too

mg2239 (changes requested):

mg2239:
fixed some issues i was having with the exporting so it should be working now, only thing is that when exporting on a day with no scheduled rides, this is what shows up
image
could you figure out how to either make this a blank csv, or have just the column headers but no data?

Becky-Hu-82:
> fixed some issues i was having with the exporting so it should be working now, only thing is that when exporting on a day with no scheduled rides, this is what shows up
> image
> could you figure out how to either make this a blank csv, or have just the column headers but no data?

just made some changes to solve this problem, but I can still only test the data in postman so don't know if it works

mg2239 (approved):
yep works on my end 👍

PR #127, opened 2021-03-06
Restrict sign in to cornell.edu emails

mg2239:
### Summary 

This pull request adds a hosted domain to Google sign in, restricting login only to users with Cornell emails.

### Test Plan 

Log out of the app, and click the sign in button.

### Breaking Changes  


mg2239:
oops this is actually already taken care of

PR #128, opened 2021-03-07
Current ride endpoint

mg2239:
### Summary 

This pull request implements a current ride endpoint for Riders, fetching their ride that's currently occurring or will occur in the next 30 minutes. If no ride matches the query, an empty JSON object is returned.

### Test Plan 

Create a ride scheduled to start within the next 30 minutes, and call the endpoint with the ride's rider.

### Breaking Changes  

### DIFF
--- a/server/router/rider.ts
+++ b/server/router/rider.ts
@@ -1,10 +1,12 @@
 import express from 'express';
 import { v4 as uuid } from 'uuid';
 import { Condition } from 'dynamoose';
+import moment from 'moment';
 import * as db from './common';
 import { Rider, RiderType } from '../models/rider';
 import { Location } from '../models/location';
 import { createKeys, validateUser } from '../util';
+import { Ride, Type } from '../models/ride';
 
 const router = express.Router();
 const tableName = 'Riders';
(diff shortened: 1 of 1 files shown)


server/router/rider.ts
```diff
@@ -60,6 +62,24 @@ router.get('/:id/favorites', validateUser('User'), (req, res) => {
   });
 });
 
+// Get current/soonest ride (within next 30 min) of rider, if exists
+router.get('/:id/currentride', validateUser('Rider'), (req, res) => {
+  const { params: { id } } = req;
+  db.getById(res, Rider, id, tableName, () => {
+    const now = moment().toISOString();
+    const end = moment().add(30, 'minutes').toISOString();
```
simranp723, line 70: It might be better style to make a separate variable for the start window instead of hardcoding "30" in case it is changed later, but since it is commented that it's within the next 30 min it is probably fine

simranp723, line 70: Also, the "end" const might get confused with the "endTime" ride field, so renaming it to "windowEnd" or something might help avoid confusion

simranp723 (approved):

PR #129, opened 2021-03-12
Student past rides

lsizemore8:
### Summary 

- Added the ability to see the past rides for a student. 
- Created a past rides component given the similarities between the driver detail and rider detail. Given the last minute changes of designs that removed name this ended up being slightly more complicated than planned but this should generally keep redundancy of code low. 
- Fixed issue with past rides for a driver where the data was not being fetched. 

- [x] implemented past rides for students
- [x] fixed past rides for driver

### Test Plan 

Past rides for student
student_past_rides_mar_11

Past Rides for Student with Date and Time
student_past_rides_with_date

Past rides for driver: 
driver_past_rides_screenshots

### Notes 

Note: the CSS is not completely correct but fixing the CSS will break other tables so I am holding off for now. 

### Breaking Changes  


mg2239 (changes requested):
looks good! i know its not on the designs, but could you add a date column to the riders too? if you can ask designers about it that would be good, but i dont really know why theres only a time column but no date lol

mg2239 (approved):
looks good! just noticed that rides are using the outdated rider type which is why accessibility isn't rendering correctly, but i think switching it now would probably break some components so need to keep track of that for later

lsizemore8:
> looks good! just noticed that rides are using the outdated rider type which is why accessibility isn't rendering correctly, but i think switching it now would probably break some components so need to keep track of that for later

So sadly I won't be able to merge this in without fixing that because the ridersTable got switched to use NewRider in the ridersContext component since keeping it as a Rider broke it with RidersTable (I don't know why).

PR #131, opened 2021-03-12
Modify location schema and how custom locations are stored in rides

mg2239:
### Summary 

This pull request utilizes Dynamoose's multiple type attribute feature to modify the Ride schema. `startLocation` and `endLocation` are now of type `String | Object`, where it is `String` if the location is an id, and `Object` if the location is a custom location. Custom locations are of the format: 
```
{
  name: string,
  address: string,
  tag: string
}
```
This enables custom location data to be stored with a ride, but not be a location in the table. No changes are made to creating a ride with a custom location. In addition, the street address w/out city and state info is now the name of the custom location, rather than Custom.

### Test Plan 

Create a ride with a custom address. The custom location data should display as normal.

### Breaking Changes  

Locations now have an `info` field (but not custom locations, possibly adding in the future), which specifies where a driver should pickup or dropoff a rider.

Custom locations are missing the id field, so on the rare chance that the location needs to be updated, you have to edit the ride instead. Also since custom locations are now no longer stored in the database, all current custom locations should be removed, which may break many rides.


mg2239:
not totally done yet

simranp723 (approved):

PR #132, opened 2021-03-13
Add scheduled query

mg2239:
### Summary 

This pull request creates a scheduled query for `/rides`, which returns all rides with `type: active` or `type: past`. The new query is also integrated with the Schedule component, which should make retrieving rides for the schedule a bit faster.
### Breaking Changes  

### DIFF
--- a/frontend/src/components/Schedule/Schedule.tsx
+++ b/frontend/src/components/Schedule/Schedule.tsx
@@ -52,20 +52,18 @@ const Schedule = () => {
 
   useEffect(() => {
     const today = moment(curDate).format('YYYY-MM-DD');
-    fetch(`/api/rides?date=${today}`, withDefaults())
+    fetch(`/api/rides?date=${today}&scheduled=true`, withDefaults())
       .then((res) => res.json())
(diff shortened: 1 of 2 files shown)


lsizemore8 (changes requested):
I believe this has breaking changes, although I have not figured out why yet. If I go to March 11, 2021 on the schedule, I get an error of "Unhandled Rejection (TypeError): Cannot read property 'id' of undefined". The stack trace will eventually lead down to the changed code in schedule.tsx. I'm not getting the same error on master, which leads me to believe it's this change.

lsizemore8:
> I believe this has breaking changes, although I have not figured out why yet. If I go to March 11, 2021 on the schedule, I get an error of "Unhandled Rejection (TypeError): Cannot read property 'id' of undefined". The stack trace will eventually lead down to the changed code in schedule.tsx. I'm not getting the same error on master, which leads me to believe it's this change.

Nvm I forgot to change the proxy, this all looks good to me!

lsizemore8 (approved):
Looks good, thanks!

PR #133, opened 2021-03-14
Integrate driver context

lsizemore8:
### Summary 

This pull request created a context for getting all of the drivers' information and integrated it into the codebase. Unlike riders, I did this all in one PR since the driver context and rider context were identical and doing this in one go seemed simpler. 
- [x] implemented driver context
- [ ] fixed Y

### Test Plan 

Tested by adding a new driver, Test McTest, and checking on all pages where driver context was added that Test McTest was there. 

Add a ride modal: 
ride_modal_mar_14

Driver Cards
screenshot_adding_driver_mar_4

Schedule
mar_14_schedule

Export
export_mar_14

### Breaking Changes  


frontend/src/components/DriverCards/DriverCards.tsx
```diff
@@ -109,6 +103,11 @@ const DriverCards = () => {
       ))}
     </div>
   );
-};
+  };
+const DriverCards = () => (
+  <DriversProvider>
+    <Drivers/>
+  </DriversProvider>
```
mg2239, line 110: instead of putting the provider around every component that uses the context, you can wrap the entire app with the provider in `App.tsx`, which removes the need to create/rename existing components

mg2239 (changes requested):

mg2239 (approved):
made it so the context refreshes after adding a driver through the modal and it works 👍 nice work

PR #136, opened 2021-03-15
Fix ride modal and assign driver modal

mg2239:
### Summary 

This pull request fixes the modal to add a ride. A data flow issue has been fixed, and rides added can now be unscheduled or active (depending on if driver is supplied). Invalid fields also now have error messages. Rides also cannot be scheduled in the past.

In addition, the assign a driver modal has been updated to better match current designs.

### Test Plan 

image
image
image
image
image

### Breaking Changes  


frontend/src/components/RideModal/Pages/RideTimes.tsx
```diff
@@ -47,6 +52,12 @@ const RideTimesPage = ({ formData, onSubmit }: ModalPageProps) => {
               },
             })}
           />
+          {errors.dropoffTime?.type === 'required' && (
+            <p className={styles.error}>Please enter a time</p>
+          )}
+          {errors.dropoffTime?.type === 'validate' && (
+            <p className={styles.error}>Invalid time</p>
```
Becky-Hu-82, line 59: Should we also add another error that indicates if the entered time is in the past (so only allow scheduling a ride in the future)?

Becky-Hu-82 (dismissed):
Just added an edge case for the modal data. The rest of the code is all good except the added ride wouldn't show up until after reloading the page and I'm not sure if we need to take care of that.

mly32:
I just realized that this form has a working date & time selector. It is different from the one used in the mini-cal. Perhaps we should combine the two in a later PR for consistency.

The modals also have different closing buttons & margins so I'll try to update them in a different PR to be consistent with designs.

In the 2nd slide to select available drivers, there is no way to cancel selecting a driver other than clicking the back button. Perhaps add another radio button for `NO DRIVER YET` or something. I'm also curious about how the screen would look like as the number of drivers increase. Maybe limit the number of drivers per line and maybe it multiple lines.


The form allows the ride to start and end in the same location. This probably isn't intended and should result in a form error.

frontend/src/components/RideModal/RideModal.tsx
```diff
@@ -60,6 +68,7 @@ const RideModal = () => {
         title={['Add a Ride', 'Available Drivers', 'Add a Ride']}
```
mly32, line 68: I just realized that this form has a working date & time selector. It is different from the one used in the mini-cal. Perhaps we should combine the two in a later PR for consistency.

The modals also have different closing buttons & margins so I'll try to update them in a different PR to be consistent with designs.

In the 2nd slide to select available drivers, there is no way to cancel selecting a driver other than clicking the back button. Perhaps add another radio button for NO DRIVER YET or something. I'm also curious about how the screen would look like as the number of drivers increase. Maybe limit the number of drivers per line and maybe it multiple lines.

The form allows the ride to start and end in the same location. This probably isn't intended and should result in a form error.

mly32 (dismissed):

frontend/src/components/RideModal/RideModal.tsx
mg2239: yeah, i just realized that the minical date selector is supposed to integrate with the add a ride modal lol, ill make a PR to integrate the two after merging this one in so we won't need the date input anymore

frontend/src/components/RideModal/Pages/RideTimes.tsx
mg2239: yeah we'd probably want that, good find

frontend/src/components/RideModal/RideModal.tsx
```diff
@@ -1,16 +1,20 @@
 import React, { useEffect, useState } from 'react';
```
mly32, line 1: image
The radio buttons would be better column aligned, (like a table or something) but that's not a big issue.

Clicking next does not display a warning/an error.

mly32: I had to resolve a merge conflict in order to test the code.

It may be worth doing a quick check on the merge conflicts. Hopefully, I did not break anything in resolving the conflict.

Becky-Hu-82: nits but could also increase the margin between the radio buttons and the two buttons in the bottom. Also if the drivers page is no longer optional do we still need the unscheduled rides table?

frontend/src/components/RideModal/ridemodal.module.css
```diff
@@ -4,6 +4,11 @@
   align-items: center;
 }
 
+.error {
+  color: #FF647C;
```
lsizemore8, line 8: Make sure you use #EB0023 for error colors instead of this, the #FF647C is too light against a white background

helenxuyang (approved):
Tested on admin web for both creating and editing rides + assigning driver, LGTM

PR #141, opened 2021-03-16
Integrate rider context

lsizemore8:
### Summary 

This integrates the rider context file with the rest of the codebase. Moreover, it moves the ridersContext to the context folder and updates it to use the NewRider type instead

- [x] implemented Rider context integration
- [ ] fixed Y

### Test Plan 
Note: adding a rider does not work due to issues I believe that is an error outside of my code to my understanding. Due to this, however, I have been unable to fully test this if the riders list is changed. I believe it is correct though since driver works as expected and driver and rider context are near identical 

riders_table_mar_15

### Breaking Changes  


mg2239 (approved):
works well on my end 👍 , made a few style changes + add refresh after adding from modal like in the driver PR. since the context design is pretty much the same id expect it to work once the modal gets fixed up

PR #142, opened 2021-03-16
Riders end date field

Becky-Hu-82:
### Summary 

This pull request adds an `endDate` field to Riders on both the frontend and the backend.
image

### Test Plan 

For now we can only go on the riders page and check the duration cell. However, because the riders added in the past didn't have the endDate field it's hard to see if the code works. In my next PR I will fix the issues with the rider modal so that we can add new riders with endDate and do full testing.


mg2239 (commented):
you have commits from the edit a ride PR on here, could you separate them?

Becky-Hu-82:
I accidentally brought changes from the other PR when creating this PR. I have just created a new one with the only changes on endDate.

PR #147, opened 2021-03-21
fixed sidebare page icon bug

aaronkang7:
This pull request fixed a bug regarding sidebar icon selection with corresponding style change.

### Test Plan 
Screen Shot 2021-03-21 at 1 33 05 AM

### DIFF
--- a/frontend/src/components/Sidebar/sidebar.module.css
+++ b/frontend/src/components/Sidebar/sidebar.module.css
@@ -22,7 +22,8 @@
 
 .icon {
   cursor: pointer;
-  margin: 1.75rem 1.625rem;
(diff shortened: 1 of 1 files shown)


frontend/src/components/Sidebar/sidebar.module.css
```diff
@@ -44,15 +45,15 @@
   background-color: #d3d2d2;
 }
 
-.circle > img {
+.circle>img {
```
mg2239, line 48: no need to make style changes if they arent relevant to the PR

mg2239 (changes requested):
works well, but i was thinking more like the user would only be able to click on the circle to switch pages. could you do that instead?

PR #148, opened 2021-03-21
Sidebar login (static)

aaronkang7:
This pull request adds static login structure in sidebar

### Test Plan 
Screen Shot 2021-03-21 at 3 05 21 AM

### Notes 
Static image has been placed as a dummy user profile picture. The Login button is a Link to the homepage for now.

### DIFF
--- a/frontend/src/components/Sidebar/Sidebar.tsx
+++ b/frontend/src/components/Sidebar/Sidebar.tsx
@@ -39,6 +39,10 @@ const Sidebar: FunctionComponent = ({ children }) => {
             </div>
           </Link>
         ))}
+        <div className={styles.logout}>
+          <img alt="dummy_profile" src="https://sincerelyvictoriat.com/wp-content/uploads/2019/11/round-profile-picture-1.png" />
+          <Link to="">Logout</Link>
+        </div>
       </div>
 
       <div className={styles.content}>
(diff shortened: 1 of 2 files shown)


mg2239:
when adding features, we want to separate them into their own PRs so the code diff doesn't get big. it looks like you have the sidebar navigation fix on this branch, and even though its small we want to remove it from this PR

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -39,6 +39,10 @@ const Sidebar: FunctionComponent = ({ children }) => {
             </div>
           </Link>
         ))}
+        <div className={styles.logout}>
+          <img alt="dummy_profile" src="https://sincerelyvictoriat.com/wp-content/uploads/2019/11/round-profile-picture-1.png" />
+          <Link to="">Logout</Link>
```
mg2239, line 44: we actually have a logout function, it can be accessed by calling useContext (https://reactjs.org/docs/hooks-reference.html#usecontext) on the context in `context/auth.ts`.

for actually showing profile pics, i dont think it should interfere with the main profile pic PR thats still in progress so you can implement that too. it'll require some modification of the AuthManager and the context i mentioned before since we need to get the user's id, so if you dont have experience with contexts feel free to dm me and i can explain further. for the dummy pfp you can just do a gray circle too

mg2239 (changes requested):

PR #150, opened 2021-03-25
Repeating ride generation function

simranp723:
### Summary 

This PR implements the functionality to generate instances of repeating rides according to the master repeating rides. It utilizes node-schedule to run this function every day to generate instances for the next day. It also uses the repeating ride's edits field to ensure that instances that have been deleted or replaced by an edit are not created.

### Test Plan 

Create a repeating ride that includes tomorrow's day, with edits that include instances for tomorrow's date.

### Notes 
- It is scheduled to run at 10:05:00am every day
- A repeating ride instance is not scheduled if there exists an edit with (delete = true) for that date
- Edit instances with (delete = true) are deleted from the DB, but the (delete = false) instances are kept and should be used as valid rides 
- The ids of edits that have been processed are removed from the master repeating ride's edits field


server/util/repeatingRide.ts
```diff
@@ -0,0 +1,115 @@
...
+    .contains(tomorrowDay);
+
+  Ride.scan(condition).exec((_, data) => {
+    data?.forEach(async (masterRide) => {
+      const {
+        id, rider, driver, startLocation, endLocation, startTime, endTime, endDate, edits,
+      } = masterRide.toJSON() as RideType;
+
+      // only continue if the endDate has not passed
+      if (endDate && endDate >= tomorrowDateString) {
```
mg2239, line 38: endDate should be formatted into YYYY-MM-DD too (or changed to be stored as such), or else some comparisons will be incorrect (ex. 01/01/2021 < 01/02/2020)

server/util/repeatingRide.ts
```diff
@@ -0,0 +1,115 @@
...
+      if (editDateOnly === tomorrowDateOnly) {
+        seenEdits.push(editId);
+        // if deleted = true
+        if (editRide.deleted) {
+          // create repeating ride instance only if it does NOT match the delete edit
+          if (!(repeatingRide.startTime === editRide.startTime
+            && repeatingRide.endTime === editRide.endTime
+            && repeatingRide.rider.id === editRide.rider.id
+            && repeatingRide.startLocation.id === editRide.startLocation.id
+            && repeatingRide.endLocation.id === editRide.endLocation.id)) {
```
mg2239, line 95: when would a repeating ride not match the delete edit?

mg2239 (changes requested):
tested somewhat lightly on my end but seems to work well and logic makes sense, just a couple comments

server/util/repeatingRide.ts
simranp723: just formatted it to YYYY-MM-DD

server/util/repeatingRide.ts
simranp723: that would be if an instance is edited and then the master ride is later edited. Then the delete edit won't match the master ride, and I think from the frontend the fake instance would show up for that day, so they would have to delete that instance again for it to be removed

mg2239: just talked with nina, if the master ride is changed after an edit is made, the edit should be preserved and a new repeating ride shouldnt be created

mg2239 (approved):

PR #155, opened 2021-03-27
Add employee

pihucode:
### Summary
This pull request allows admins to be created through the Employees page. 

- [x] Added role selector to the "Add an employee" modal.
- [x] Removed vehicle field.
- [x] Renamed driver components (i.e, DriverCards -> EmployeeCards).
- [x] Refactored `DriversProvider` to `EmployeesProvider`.
- [x] Added `Admin` and `Employee` types.
- [x] Changed path from `/drivers` to `/employees`.

### Test Plan
Created user "Test Admin" that has only the admin role and the user "Test Both" that has both roles driver and admin. 

### Notes 
Conditionally rendering the availability element results in an error:
`Type 'Element | null' is not assignable to type 'Element'. Type 'null' is not assignable to type 'Element'.`
Since admins do not have availability info, the text is changed to "N/A" as a temporary fix.


mg2239 (approved):
works very well, nice work!

PR #156, opened 2021-03-27
Driver error messages

lsizemore8:
### Summary 

This PR added error messages to the driver modal. The location and styling was based off the current PR for the ride modal. The sizes for this modal were adjusted slightly to prevent the driver modal from increasing too much in size. 

Additionally, the repeat on buttons now indicate what button the user is on when focused on that button for accessibility purposes. 

- [x] implemented driver error messages
- [ ] fixed Y

### Test Plan 
No errors
no_errors_driver_modal_april_1

With errors
driver_modal_with_errors_apr_1

### Notes 

This is not a perfect fix. If you correctly enter the startTime but not the endTime (or vice versa) the spacing will appear off. This is due to how the flexbox is organized & how the error messages were positioned. If I have time I will make a separate PR fixing this but it's minor enough to be okay for now. 

Additionally, this will break if the screen size is reduced too drastically. This is an issue with the base modal class and should be fixed in a separate PR. 
### Breaking Changes  


mly32 (changes requested):
There are some issues with the modal at small screen widths. I'm not sure if this is a big issue.
image


Not sure what is causing this one:
When I press the add a driver I get the yellow popup but there is no red error below it.

image
image

The error for working hours is incorrect here:

image


It is possible to add a driver without selecting any of the days for working hours.

mg2239:
> There are some issues with the modal at small screen widths. I'm not sure if this is a big issue that should be fixed in this pr.

all the admins will use the site on desktop, so its not a big issue rn

frontend/src/components/EmployeeModal/EmployeeInfo.tsx
```diff
@@ -1,45 +1,58 @@
...
+      <div className = {styles.col2}>
+        <Input
+          name='netid'
+          type='text'
+          placeholder='NetID'
+          className={cn(styles.input)}
+          ref={register({ required: "Please enter a valid NetID" })}
+        />
+        {errors.netid && <p className={styles.error}>{errors.netid.message}</p>}
+      </div>
```
mly32, line 32: image
image

There are some weird artifacts with the input sizing (maybe show some borders to see where the issue is). I think the email input is too long.


I'm not sure if this is the intended design, but the contrast is bad for the day when I click on it.
image

It probably has to do with focus coloring since when I click elsewhere I get.
image

lsizemore8:
Closed due to the License changes

PR #157, opened 2021-03-28
web notifications

mly32:
### Summary 

Added a service worker and updated the backend to handle web notifications.

For now, the backend uses a set containing subscriptions and the frontend subscribes to get global notifications when they head to the notifications page.

image

### Test Plan 

Change the frontend proxy to `http://localhost:3001`

head to `http://localhost:3000/dashboard/notifications`

View notifications

There is a test endpoint on the backend (doesn't need auth) to send notifications.

Use `http://localhost:3001/api/notification/send` and post. Here is an example body:

```
{
    "title": "foo",
    "body": "bar"
}
```

Test these when the tab is not in focus as well.

### Notes 

I have not thoroughly tested the checks for determining whether a browser supports notifications (e.g., chrome incognito does not). 

There needs to be further testing for when a subscription expires.


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -60,7 +60,7 @@ const Schedule = () => {
       .then((res) => res.json())
       .then(({ data }) => {
         setEvents(
-          data.map((ride: Ride) => ({
+          data.filter((ride: Ride) => (ride.driver && ride.driver.id)).map((ride: Ride) => ({
```
mg2239, line 63: can remove this filter

server/app.ts
```diff
@@ -12,12 +13,14 @@ import vehicle from './router/vehicle';
...
 import upload from './router/upload';
 import auth from './router/auth';
+import notification from './router/notification';
 
 const port = process.env.PORT || 3001;
 
 dynamoose.aws.sdk.config.update(config);
 
 const app = express();
+app.use(cors());
```
mg2239, line 23: why is cors needed?

mg2239 (approved):
just a couple comments, overall works well

PR #158, opened 2021-03-29
[WIP] Profile pictures

pihucode:
### Summary
This pull request enables profile photos to be uploaded when adding users and displays profile photos for drivers and riders. 

The jpg or png photo file is converted to a base64 string that is sent to the endpoint `/api/upload` which returns an URL. This URL is then set as the `photoLink` value. 

- [x] Preview photo on upload
- [x] Upload photo in `DriverModal` and `RiderModal`
- [x] Display photo in `UserDetail` if the photo exists

### Test Plan
image
image
image

### Notes 
Only driver photos have been tested so far. New riders cannot be created so profile photos for riders cannot be currently tested.
Added `photoLink?: string` field to Driver and NewRider types.


mg2239 (approved):
nice work!

PR #159, opened 2021-03-31
Format headers consistently, add notification bell

jj534:
### Summary 

- Fix headers so that they're consistent across all pages
- Add notification bell to header

### Test Plan 

image


frontend/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import styles from './notification.module.css';
+import { notificationBadge, notificationBell } from '../../icons/other';
+
+const Card = () => (
```
mg2239, line 5: whys this called card?

frontend/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,12 @@
...
+    <div className={styles.container}>
+      <img src={notificationBell} className={styles.bell} alt="notification bell icon" />
+      <img src={notificationBadge} className={styles.badge} alt="notification badge icon" />
```
mg2239, line 8: can you add state to make this show up only when notifications are present? the notification logic doesnt need to be implemented, just the state

frontend/src/pages/Dashboard/page.module.css
```diff
@@ -14,8 +14,15 @@
...
-  margin: 3rem;
+.rightSection {
+  margin-right: 3rem;
+  display: flex;
+  align-items: center;
+}
+
+.rightSection > div {
+  margin-left: 1rem;
 }
```
mg2239, line 28: sorry if this wasn't clear, i wanted you to follow the header styles that are used for the employee and students pages in the designs
image
if this changes the homepage styles thats okay, that page will require some more work in a different PR

mg2239 (changes requested):

frontend/src/pages/Dashboard/page.module.css
jj534: do you mean I should use the existing `.margin3` css class instead of introducing new `.rightSection` css class?

if I use `.margin3`, the notification bell component doesn't render at the correct position. Would that be okay?

image

frontend/src/components/Notification/Notification.tsx
jj534: no idea. I think my editor may have auto-fixed it. will change it to `Notification`

frontend/src/components/Notification/Notification.tsx
jj534: yeap! will do

frontend/src/pages/Dashboard/page.module.css
mg2239: nah you can use the rightSection class. feel free to introduce new styles, change existing styles, and reformat any pages as you see fit, as long as all the pages align with the designs (besides homepage, but it would be good if you could structure it similar to other pages if possible)

jj534: okay sounds good. which part did you feel were different from the designs?

mg2239: mainly the padding and header font size

jj534: okay I've edited the css so that the padding and header font size are closer to the designs. let me know if there's anything else that you want fixed

mg2239 (approved):

PR #161, opened 2021-04-02
Fix Auth flow vulnerability

zack-ashen:
### Summary 

Fixed auth flow vulnerability where an email could be spoofed from the frontend to login as an admin. Removed email from post request to backend and instead request email from LoginTicket payload.

- [x] User email is not sent from frontend
- [x] Email is received from LoginTicket payload


mg2239:
separate the collapsible component commits and the auth flow commits to different PRs

mg2239 (approved):
looks good, nice work!

PR #162, opened 2021-04-02
Filter delete rides

jj534:
### Summary 

This pull request is filters deleted rides regardless of the requested query. Deleted rides will no longer show up on the frontend.


mg2239 (approved):

mg2239 (changes requested):
actually I just remembered why we don't filter out deleted rides, we need them for the rider app. so the filtering should be on the frontend at least for now

PR #164, opened 2021-04-02
Sort employees, students, locations alphabetically

jj534:
### Summary 

This pull request sorts employees, students, locations alphabetically

### Test Plan 
image
image
image

### DIFF
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -115,11 +115,6 @@ const EmployeeCards = () => {
   const { admins, drivers } = useEmployees();
 
   const allDrivers = [...admins, ...drivers];
-  allDrivers.sort((a: Driver | Admin, b: Driver | Admin) => {
-    if (a.firstName < b.firstName) { return -1; }
-    if (a.firstName > b.firstName) { return 1; }
-    return 0;
-  });
 
   return (
     <div className={styles.cardsContainer}>
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -115,6 +115,11 @@ const EmployeeCards = () => {
   const { admins, drivers } = useEmployees();
 
   const allDrivers = [...admins, ...drivers];
+  allDrivers.sort((a: Admin, b: Admin) => {
+    if (a.firstName < b.firstName) { return -1; }
+    if (a.firstName > b.firstName) { return 1; }
+    return 0;
+  });
 
   return (
     <div className={styles.cardsContainer}>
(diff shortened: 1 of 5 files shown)


frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -103,25 +103,26 @@ const EmployeeCard = ({
...
   );
 };
+
 const EmployeeCards = () => {
   const { admins, drivers } = useEmployees();
 
+  const allDrivers = [...admins, ...drivers];
+  allDrivers.sort((a: Driver | Admin, b: Driver | Admin) => {
+    if (a.firstName < b.firstName) { return -1; }
+    if (a.firstName > b.firstName) { return 1; }
```
mg2239, line 113: maybe do it by full name instead, and move sorting into the EmployeeContext so its consistent throughout the site

frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -1,10 +1,17 @@
...
 import { Row, Table } from '../TableComponents/TableComponents';
 import { useRiders } from '../../context/RidersContext';
 
 const StudentsTable = () => {
   const { riders } = useRiders();
+  riders.sort((a: NewRider, b: NewRider) => {
+    if (a.firstName < b.firstName) { return -1; }
+    if (a.firstName > b.firstName) { return 1; }
+    return 0;
+  });
```
mg2239, line 13: same thing here

mg2239 (changes requested):

frontend/src/components/EmployeeCards/EmployeeCards.tsx
jj534: 1. maybe do it by full name instead
I don't think there's a fullName attribute for employee and driver objects. Could you clarify what you mean?

2. move sorting into the EmployeeContext so its consistent throughout the site
I added the alphabetically sorting logic to EmployeeContext, but since EmployeeCards merges admins and drivers into a single allDrivers array, I kept the allDrivers.sort to sort the merged array alphabetically

mg2239: fullName would be ``${a.firstName} ${a.lastName}``

frontend/src/components/UserTables/StudentsTable.tsx
mg2239: should also be moved to RidersContext

mg2239 (changes requested):
make sure it's non case sensitive too, getting this on the students table
image

frontend/src/components/EmployeeCards/EmployeeCards.tsx
jj534: wouldn't `${a.firstName} ${a.lastName}` just sort by `a.firstName` ?

mg2239: if there are two people with the same first names, then we'd need to look at the last name

mg2239 (approved):

PR #165, opened 2021-04-05
Implement get all repeating rides endpoint

mg2239:
### Summary 

This pull request creates a get all repeating rides endpoint, to help retrieve master repeating rides that have already occurred, but are still repeating.

### Breaking Changes  

The `endDate` field is now formatted as YYYY-MM-DD, to fix bugs when comparing times.

### DIFF
--- a/server/router/ride.ts
+++ b/server/router/ride.ts
@@ -55,9 +55,7 @@ router.get('/repeating', validateUser('User'), (req, res) => {
   let condition = new Condition('recurring')
     .eq(true)
     .where('endDate')
-    .ge(now.substring(0, 10))
-    .where('startTime')
-    .le(now);
+    .ge(now.substring(0, 10));
   if (rider) {
     condition = condition.where('rider').eq(rider);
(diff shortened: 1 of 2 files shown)


server/router/ride.ts
```diff
@@ -48,6 +48,22 @@ router.get('/download', (req, res) => {
...
 
+// Get and query all master repeating rides in table
+router.get('/repeating', validateUser('User'), (req, res) => {
+  const { query: { rider } } = req;
+  const now = moment().toISOString();
+  let condition = new Condition('recurring')
+    .eq(true)
+    .where('endDate')
+    .ge(now.substring(0, 10))
+    .where('startTime')
```
helenxuyang: (Mentioned on Slack) I think the startTime check would exclude repeating rides that start in the future

helenxuyang (approved):
I believe this looks good aside from the startTime filter, just approving because that's a quick fix and should be able to merge afterwards!

PR #167, opened 2021-04-07
implemented toast notifications for Rider and EmployeeModal

aaronkang7:
### Summary 

This pull request implements toast notifications for rider and employee modals. There seems to be a problem with +Add Ride submission, but if that is fixed, the toast should work as well. The add Location form doesn't seem to have a separate modal yet so I didn't integrate the toast notification to it just yet.  

Toast notifications are automatically dismissed after 4 seconds after form submision.

- [x] implemented X
- [ ] fixed Y

### Breaking Changes  


frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -62,10 +67,12 @@ const EmployeeModal = () => {
       })).then(() => refreshDrivers());
     }
     closeModal();
+    setToast(true);
   };
 
   return (
     <>
+      {showingToast ? showToast('The employee has been added.') : null}
```
mg2239, line 75: you can just use the component itself here

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -62,10 +67,12 @@ const EmployeeModal = () => {
       })).then(() => refreshDrivers());
     }
     closeModal();
+    setToast(true);
```
mg2239, line 70: we dont want this to show if they close out before submitting, only when a submission is successful

mg2239 (changes requested):
made a few small changes cause i thought it was good to go, but noticed on the designs that the toast also shows up when the schedule is downloaded. could you add that in too?

aaronkang7:
> made a few small changes cause i thought it was good to go, but noticed on the designs that the toast also shows up when the schedule is downloaded. could you add that in too?

mg2239 (changes requested):
can you use the download icon too

mg2239 (approved):
oops i was looking at the wrong design, nvm looks good lol

aaronkang7:
> oops i was looking at the wrong design, nvm looks good lol

oh, I'll just delete the commit then!

PR #169, opened 2021-04-09
Driver error messages fixed

lsizemore8:
### Summary 

Given that the driver-error-messages (#156) accidentally got a technically changed version of the LICENSE file on it and could not be reverted, this is a copy of that PR into a new PR. 

- [x] implemented driver error messages
- [ ] fixed Y

### Test Plan 
See #156. Tested after the swap happened to ensure no other changes occurred in the swap. I fully copied the EmployeeModel folder from #156 to here so no errors should have occurred that were not in #156 regardless

### Notes 

If the dev leads read this, I acknowledge that the first commit name is pretty bad. This was due to a commit (not a push) on master of the license to unblock myself from committing or making new branches. When I created the new branch, the old commit came along with the branch. I frankly did not know one could do this until today and is an interesting, but not great lesson learned. That being said, if this triggers the poor commit name dock, I fully accept that. 

### Breaking Changes  


mly32 (changes requested):
This took forever to check correctly, but the padding for email and phone number is different. When you click on each input, they are not horizontally aligned.

image
It becomes very apparent when all the errors are also visible (it may be nice for every error line to be aligned).

Design thing: there plus on the right should also act as an upload a picture button, but if it differs from the designs, no need to change it.

A phone number of `11111111111111111111111111111111111111111111111111111111111111111111111111111` is correct when it probably shouldn't be.


The error where the times are equal does not pass 
image
which is good. However, the error does not display an error message.


It would be nice for the time select widths to be equal.
image

mg2239 (approved):

PR #171, opened 2021-04-11
Repeating ride creation support

mg2239:
### Summary 

This pull request adds the ability to create repeating rides using the POST /rides endpoint. To create a repeating ride, the request must contain the recurring, recurringDays, and endDate fields. In addition, the requestedEndTime field has been removed, as it is not needed anymore.

### Breaking Changes  

If the frontend uses requestedEndTime, any references must be replaced with endTime.


helenxuyang (approved):
Everything looks good!

PR #173, opened 2021-04-11
added notifications table

mly32:
### Summary 

Added notifications table.

image

image

image

### Test Plan 

Try running on different pages.

### Breaking Changes  


frontend/src/components/Notification/Notification.tsx
```diff
@@ -1,18 +1,84 @@
...
+
+  const onClick = () => {
+    console.log('clik');
+  };
+
+  return (
+    <Popup
+      trigger={
+        <button className={styles.bell} onClick={onClick}>
+          <span className={styles.span} />
```
mg2239, line 51: what are the spans for?

frontend/src/components/Notification/Notification.tsx
```diff
@@ -1,18 +1,84 @@
...
-    )}
-  </div>
-);
+type Message = {
+  time: Date;
+  title: string;
+  body: string;
+};
+
+const Notification = () => {
```
mg2239, line 15: the notification components is dismounted and remounted on page change, which resubscribes the user. ideally wed want them to only subscribe once, would it be possible to either create a shared header component across routes, or move notification state to a context?

mg2239 (changes requested):
looks good! just a couple changes mentioned in the comments, and also id like the notifications to be better aligned to styles, such as profile picture (can be placeholder for now), date formatting as month day rather than mm/dd/yyyy, only showing one line for notifications, etc. 

for the clickable "view" link you dont need to implement functionality right now, but once the ride info can be sent through notifications it should navigate to the schedule page and change the schedule date to the ride date

frontend/src/components/Notification/Notification.tsx
mly32: Right, this is pretty important. I'll implement this in a different PR.

frontend/src/components/Notification/Notification.tsx
mly32: They were for spacing but are no longer necessary. I have removed them.

mg2239 (approved):
nice work!

PR #175, opened 2021-04-12
implemented tag component

aaronkang7:
### Summary 

This pull request implements address tag components. Color for corresponding location tag still needs to be confirmed by designers. 

- [x] implemented tag feature (made separate component)
- [x] Color hex codes confirmed by designers.

### Breaking Changes  


mg2239 (changes requested):
tags look good! but by component i meant a separate component, since we also use the tags when displaying ride information on the schedule. could you separate the code out from the table cells? you dont need to add tags to the tables either, since we're reimplementing all the tables this sprint

erika should also be getting back to you about the colors to use, so be sure to add those too 👍

frontend/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+    case ('custom'): {
+      tagStyle = styles.custom;
+      break;
+    }
+    default: {
+      tagStyle = styles.inactive;
+      break;
+    }
+  }
```
mg2239, line 42: why not styles[tagStyle]?

mg2239 (changes requested):
make sure the right tag styles are being used in the right places, and make sure the tags are positioned correctly

ex. should be not reduced, tag circles should be vertically centered
image

frontend/src/components/Tag/Tag.tsx
aaronkang7: oh... i didnt know you could do that ahah

mg2239 (changes requested):
looks good overall, the tag positioning can probably wait until the new table style are merged in. please implement the new tag colors though

aaronkang7:
> looks good overall, the tag positioning can probably wait until the new table style are merged in. please implement the new tag colors though

new tag colors? I believe the colors were confirmed by designers quite a while ago. Were there any design changes that I may not be aware of? 

I found the hex codes from the "Design System" page of the Figma

mg2239 (approved):
ah yeah i see now, the full sized tag styles weren't implemented correctly so thats why i couldnt see the new colors. just fixed and should be good to go

PR #176, opened 2021-04-12
endpoint for repeating ride edits

simranp723:
### Summary 

This PR adds an endpoint that creates edit instances for a repeating ride and updates its edits field. 

First the date of the edit is added to the master ride's deleted field. If the instance is being edited, a replace edit instance with is created with the passed in field values.

### Test Plan 

Test using `PUT /api/rides/:id/edits`:
- id = id of the master repeating ride
- deleteOnly = true if an instance is being deleted, false if it is being edited/replaced
- origDate = date of the original ride instance being deleted/edited, in ('YYYY-MM-DD') format
- startTime, endTime, startLocation, endLocation = values to be used for the replace edit instance


server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+      // if deleteOnly = false, create a one-time ride with the new fields
+      if (!deleteOnly) {
+        const newRide = new Ride({
+          id: uuid(),
+          rider,
+          startLocation,
+          endLocation,
+          startTime,
+          endTime,
+          driver,
```
mg2239, line 190: a few things:
 - can just use master ride's rider
 - edits are unscheduled so they dont need a driver
 - if one of those fields aren't provided, they should use the value in the master ride

server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+          deleted: false,
+        });
+        db.create(res, replaceRide);
+        newEdits.push(replaceId);
+      }
+      const operation = { $SET: { edits: newEdits } };
+      db.update(res, Ride, { id }, operation, tableName);
+    }
```
mg2239, line 231: include a check if origDate < masterFirstDay

server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+          startTime,
+          endTime,
+          driver,
+          deleted: false,
+        });
+        db.create(res, replaceRide);
+        newEdits.push(replaceId);
+      }
+      const operation = { $SET: { edits: newEdits } };
+      db.update(res, Ride, { id }, operation, tableName);
```
mg2239, line 230: we should only update edits if we know the edits were made successfully

server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+      // if deleteOnly = false, create a replace edit with the new fields
+      if (!deleteOnly) {
+        const replaceId = uuid();
+        const replaceRide = new Ride({
+          id: replaceId,
+          rider,
+          startLocation,
+          endLocation,
+          startTime,
+          endTime,
```
mg2239, line 222: same thing w/ using master ride fields if not provided, but if the edit is just to update startLocation for example, we would still need to adjust startTime and endTime to origDate

server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+        endTime: origEndTime,
+        driver: masterRide.driver,
+        deleted: true,
+      });
+      db.create(res, deleteRide);
+
+      const newEdits = [deleteId];
+      if (masterRide.edits) {
+        newEdits.concat(masterRide.edits);
+      }
```
mg2239, line 211: edits is included by default on all repeating rides now

server/router/ride.ts
```diff
@@ -145,6 +145,93 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+        });
+        db.create(res, newRide);
+      }
+    } else {
+      const deleteId = uuid();
+      const deleteRide = new Ride({
+        id: deleteId,
+        rider: masterRide.rider,
+        startLocation: masterRide.startLocation,
+        endLocation: masterRide.endLocation,
```
mg2239, line 200: make sure if startLocation is a preset location, youre only passing the id into startLocation (and same for endLocation)

mg2239 (changes requested):

mg2239 (approved):
had to move lines around to fix some issues with the endpoint trying to send multiple responses, but looks good!

PR #177, opened 2021-04-12
fix rider type, implement new modal, code cleanup

Becky-Hu-82:
### Summary 

This pull request fixes issues for adding a new rider.

- Fixed the rider type on the frontend
- Cleaned up / fix some old code based off of the new rider type
- Implemented the new design for the rider modal
- Added validator on the modal's fields

### Test Plan 

Play around the modal and see if all edge cases have been covered.
image
image


mg2239 (changes requested):
made some small changes to get everything working, just remove all the unused code including comments and should be all good

frontend/src/components/FormElements/formelements.module.css
```diff
@@ -16,7 +16,7 @@
 
 .timeInput,
 .dateInput {
-  font-size: 1.25rem;
+  font-size: 1rem;
```
lsizemore8, line 19: Changing the font size here makes the text in the add new ride modal seem weirdly small. It may be better to change this text in the add rider CSS module instead of changing it here

lsizemore8 (changes requested):
A few things: 
1. This is something relatively minor but I had the same issue with the driver modal so I wanted to bring it up here as well. If you look carefully, the error messages and the inputs are slightly misaligned. The issue is likely that there is some padding added to the inputs but not the error messages. If that can be aligned, that would improve the styling of it some. 
2. For accessibility purposes, all inputs need to have a corresponding <label>, but I'm adding a fix so there's always a hidden label that people can use in cases like these. I can chat with matt to see if it's worth just fixing that label as part of the accessibility fixes or if we want to merge my fix in with a different branch and you add it here
3. The inputs are overflowing into each other. Trying setting a max-width in the CSS. 
image
4. Again, pretty minor detail, but the spacing between the first date and the to vs the end date and the to seems pretty uneven. 
image

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,157 @@
...
-        <div className={cn(styles.gridR4, styles.gridC2)}>
-          <Label htmlFor="end">End Date:</Label>
-          <Input
-            type="date"
-            name="end"
-            ref={register({ required: true })}
-          />
+        <div className={cn(styles.gridR3, styles.gridCAll)}>
+          <p>Duration</p>
+          <div style={{ display: "flex" }}>
```
lsizemore8, line 118: In general, using inline styling isn't really great, just because it makes it harder to keep track of things/update/make accessible or work with media queries (the last part being the key detail here for the future of admin web). Can you move the inline styling to the css file/use classes instead?

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,157 @@
...
+                type="date"
+                name="joinDate"
+                ref={register({ required: true })}
+                className={styles.riderDate}
+              />
+              {errors.joinDate && (
+                <p className={styles.error}>Please enter a join date</p>
+              )}
+            </div>
+            <p style={{ margin: "0 1rem" }}>to</p>
```
lsizemore8, line 130: Same thing here with the inline styling

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,157 @@
...
+              )}
+              {errors.end?.type === "validate" && (
+                <p className={styles.error}>Invalid end time</p>
+              )}
+            </div>
+          </div>
         </div>
       </div>
-      <Button type="submit">Add a Rider</Button>
+      <Button type="submit">Add a Student</Button>
```
lsizemore8, line 154: Last minor fix I swear, on the designs and the employee modal (which this is v similar to) the submit button is right aligned. Generally that's better practice than center-aligned because it fits the flow of how you read forms better (according to info 1300). Unless the designers told you otherwise (in which case please ignore me) can you make the submit button right aligned instead of center aligned

frontend/src/components/Modal/RiderModal.tsx
```diff
@@ -55,37 +50,30 @@ const RiderModal = () => {
...
       setIsSubmitted(false);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [formData, isSubmitted, refreshRiders, withDefaults]);
 
   return (
     <>
-      {showingToast ? <Toast message='The student has been added.' /> : null}
-      <Button className={styles.addRiderButton} onClick={openModal}>+ Add Student</Button>
+      {showingToast ? <Toast message="The student has been added." /> : null}
```
lsizemore8, line 71: Any reason to fix double vs single quotes? I don't think you need to fix it either way, I'm more just curious.

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,157 @@
...
       endDate,
     });
   };
+
   return (
     <form onSubmit={handleSubmit(beforeSubmit)} className={styles.form}>
       <div className={cn(styles.inputContainer, styles.rideTime)}>
-        <div className={cn(styles.gridR1, styles.gridC1)}>
+        <div className={cn(styles.gridR1, styles.gridCSmall1)}>
           <Input
```
lsizemore8, line 45: For all of the inputs you'll need to add some sort of label/identifier. the SRLabel that was created should work in this case. The placeholder text is not sufficient & we'll need some sort of label (even if hidden) to make it accessible. It's tedious but unavoidable. I can give you an example of how to use it if you need.

lsizemore8 (changes requested):
It generally looks fine. I just left a couple small fixes that need to be done and then I'll approve it.

frontend/src/components/Modal/ridermodal.module.css
```diff
@@ -34,9 +63,16 @@
   grid-row: 3;
 }
 
-.gridR4 {
-  grid-row: 4;
+.error {
+  color: #ff647c;
```
lsizemore8, line 67: Still confirming with designers here but this font/font size pairing is not accessible. If I don't hear back soon I can make this fix in accessibility-fixes but I may request you to change this color because it doesn't meet contrast ratio.

lsizemore8: Update: change the color to #EB0023

frontend/src/components/Modal/RiderModal.tsx
Becky-Hu-82: i think there are some weird settings in my prettier. Just turned it off.

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,181 @@
...
+                <p className={styles.error}>Invalid end time</p>
+              )}
+            </div>
+          </div>
         </div>
       </div>
-      <Button type="submit">Add a Rider</Button>
+      {/* TODO: styling */}
+      <div className={styles.buttonContainer}>
+        <Button className={styles.cancel} outline={true} onClick={() => cancel()}>Cancel</Button>
```
lsizemore8, line 176: Right now, if you press enter regardless of where you are on the modal, it will press cancel because the default button type is "submit". Add type = "button" to this button to fix this. That also does bring up the question for the need for the cancel button given the x at the top of the modal but that's a design question, not a me question.

frontend/src/components/Modal/ridermodal.module.css
lsizemore8: Make sure you change the error color here.

lsizemore8 (commented):
Fully realized that you may not be done so I'll just leave these two comments here

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,183 @@
...
+              {errors.end?.type === "validate" && (
+                <p className={styles.error}>Invalid end time</p>
+              )}
+            </div>
+          </div>
         </div>
       </div>
-      <Button type="submit">Add a Rider</Button>
+      <div className={styles.buttonContainer}>
+        <Button type = "button" className={styles.cancel} outline={true} onClick={() => {console.log('clicked');cancel()}}>
```
lsizemore8, line 176: Is there any reason to keep the console.log here?

frontend/src/components/Modal/ridermodal.module.css
```diff
@@ -34,9 +63,44 @@
...
+.buttonContainer {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  grid-template-rows: repeat(1, 1fr);
+  column-gap: 1rem;
+  row-gap: 1rem;
+}
+
+.submit {
+  /* margin-left: 5rem; */
```
lsizemore8, line 97: Remember to remove the comments before merging this!

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -1,112 +1,183 @@
...
+                ref={register({
+                  required: true,
+                  validate: (endDate) => {
+                    const joinDate = getValues("joinDate");
+                    return joinDate < endDate;
+                  },
+                })}
+                className={styles.riderDate}
+              />
+              {errors.end?.type === "required" && (
```
lsizemore8, line 165: You'll either have to change this to errors.endDate?.type or change the name of this to 'end'. Right now the errors don't appear because there is no error for an input of end.

lsizemore8 (changes requested):
It's pretty much almost there! The end date errors don't work right now because the errors is using errors.end?.type instead of errors.endDate?.type to match the name of the input. I left a couple other small comments but once those are fixed i think this is good to go

lsizemore8 (approved):
Looks good to me!

mg2239 (approved):

PR #180, opened 2021-04-15
Edit employees

pihucode:
### Summary
This pull request enables employees to be edited through `EmployeeModal`.

- [x] Refactor `EmployeeModal` to support editing.
- [x] Support updating photos for existing employees.

Values of an existing employee account are passed into `EmployeeModal` as props. Each input component of the modal sets the prop value as their `defaultValue`.

#### Working Hours
`getAvailabilityMap` creates an availability map from `existingAvailability: string[][]` with time ranges as keys and repeating days as values. 

`availabilityMapToArray` transforms the availability map into an array which can be mapped to JSX.Elements.

`formatTime` transforms a time string into a string that can be passed into the `defaultValue` field of the `Input` component.
For strings ending in "pm", 12 should be added to the hour. For example, `"7pm"` should become `"19:00"`. The function `formatTime` handles two edge cases:
- "12pm" should remain "12:00" and not "24:00".
- "12am" should become "00:00".

#### Roles
The expected usage behaviour is for drivers to remain as `Driver` or change to `Both` and admins to remain as `Admin`.
Currently changing roles from driver to admin or vice versa is not enforced and may result in duplicate accounts being created on two different endpoints.

### Test Plan
Created driver and admin employees and edited them through the frontend.

### Notes 
The Sunday and Saturday day inputs under the "Working Hours" section seem to be broken. This was a pre-existing issue before this PR.

Existing employees who do not have a photo are able to update their photo through the edit modal. However, photos do not update for employees who have previously uploaded a photo.


mg2239 (approved):
really good work!

mg2239 (approved):

PR #183, opened 2021-04-16
Tab Switcher

pihucode:
### Summary 
This PR creates a tab switcher component that can support multiple tabs.

- [x] Create a tab switcher component

`TabSwitcher` reads the labels of its children and renders the contents of the child whose label matches the current label. The default current tab is set to the first child.

### Test Plan
image
image


aaronkang7 (commented):
Works well! But I was wondering if the notification bell is part of the analytics page? It doesn't seem to be part of the figma. If it is supposed to be there, just disregard this comment! Either way, it would be a very quick fix.

pihucode:
> Works well and code seems good! But I was wondering if the notification bell is part of the analytics page? It doesn't seem to be part of the figma. If it is supposed to be there, just disregard this comment! If not, it would be a very quick fix.

Good question! I will confirm with the designers on the notification bell.

mg2239 (approved):
im pretty sure the notifications should be accessible everywhere, so i'd say it should be okay. but feel free to wait until a designer responds to merge in

PR #185, opened 2021-04-18
Change repeating ride schema

mg2239:
### Summary 

This pull request revises the repeating ride schema to change `deleted` from a boolean to an array. Since deleted rides are only checked by date, not creating them and only storing the dates will reduce the amount of work needed for the frontend to figure out how to display repeating rides.

In addition, delete operations are fixed, and the DELETE ride endpoint now deletes all edits for a repeating ride.

### Test Plan 

Create a repeating ride, and run the createRepeatingRides function in repeatingRide.ts

### Breaking Changes  

Because this changes how repeating rides are being displayed, the frontend should account for these changes after this PR merges.

### DIFF
--- a/server/util/repeatingRide.ts
+++ b/server/util/repeatingRide.ts
@@ -49,9 +49,7 @@ function createRepeatingRides() {
         endLocation: endLocation.id ?? endLocation,
       });
 
-      ride.save()
-        .then((doc) => console.log(doc.toJSON()))
-        .catch((err) => console.log(err));
+      ride.save().catch((err) => console.log(err));
     });
   });
 }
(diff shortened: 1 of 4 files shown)


server/util/repeatingRide.ts
```diff
@@ -12,104 +11,45 @@ export default function initSchedule() {
...
     .eq(true)
-    .and()
     .where('recurringDays')
-    .contains(tomorrowDay);
+    .contains(tomorrowDay)
+    .where('endDate')
+    .ge(tomorrowDateOnly)
+    .where('deleted')
+    .not()
+    .contains(tomorrowDateOnly);
```
simranp723, line 29: i think we still need to check that the startDate is less than tomorrow's date

simranp723 (changes requested):

server/util/repeatingRide.ts
mg2239: yep good catch

simranp723 (approved):

PR #186, opened 2021-04-19
Driver join date

aaronkang7:
### Summary 

This PR adds driver join date to the frontend modal and backend schema. 

- [x] Made modal changes (add join date) to Employee modal
- [x] Made corresponding backend changes

### Notes 

Screen Shot 2021-04-19 at 2 41 53 AM
Screen Shot 2021-04-19 at 2 41 45 AM

### Breaking Changes  


mg2239:
looks good! just gonna hold off until designs are finished before approving

frontend/src/components/EmployeeModal/StartDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+import styles from './employeemodal.module.css';
+
+const StartDate = () => {
+  const { register } = useFormContext();
+
+  return (
+    <div className={cn(styles.col1, styles.workingHours)}>
+      <Label htmlFor="start">Start Date:</Label>
+      <Input
+        type="date"
```
lsizemore8, line 16: Right now the label & input are not connected, so accessibility wise the label means nothing. When you use "htmlFor", you're connecting the htmlFor to the input's id value, which you have not set here. To connect the label to the input, you need to do one of two things: 

1. Give the input type you have here an id of "start"
2. Give the label some id attribute and in the input, put in aria-labelledby={id_of_label} where id_of_label is the string id you gave to that label.

aaronkang7: Thank you for the feedback! It would be linked now :)

mg2239 (approved):
not sure when designs are gonna come in but ill approve for now

PR #190, opened 2021-04-20
Create analytics table

mg2239:
### Summary 

This pull request implements the analytics table with editing capabilities. The table works for both ride and driver data, and is compatible with the backend schema.

### Test Plan 

image
image

### Notes 

Should be connected to the backend in a separate PR.

The component currently logs the edits to the console for debugging, and on submit just clears the edits. 

### Breaking Changes  


frontend/src/components/AnalyticsTable/AnalyticsTable.tsx
```diff
@@ -0,0 +1,274 @@
...
+      initRideData(table.data);
+    } else {
+      initDriverData(table.data);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [type, drivers.length]);
+
+  return (
+    <div className={styles.tableContainer}>
+      <button
```
lsizemore8, line 212: This button is not accessible as it has no text describing what it's doing. Add an aria-label to it to make it accessible (see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute)

frontend/src/pages/Dashboard/Analytics.tsx
```diff
@@ -1,14 +1,15 @@
 import React from 'react';
+import AnalyticsTable from '../../components/AnalyticsTable/AnalyticsTable';
 import TabSwitcher, { Tab } from '../../components/TabSwitcher/TabSwitcher';
 
 const Analytics = () => (
    <div>
       <TabSwitcher>
```
lsizemore8, line 7: I'm not sure if this should be fixed here or in accessibility-fixes, but this tabswitcher is not accessible. Because it uses <h1>, it's not focusable by keyboard/screen reader will treat it like normal text instead of something interactable. This means if I try to use the keyboard to navigate to here, I can't reach it.

lsizemore8 (changes requested):
Generally this seems to work fine (minus the fact the backend isn't connected), but left a couple accessibility issues I noticed in the comments. One of the issues we may not need to fix in this PR, the other definitely needs to be fixed here & should be a quick fix. 

Moreover, I'm not sure if this is meant to be responsive, but if it is, the edit button has an absolute positioning so it stays in the middle when you scroll right & the window is as half size 
image

frontend/src/pages/Dashboard/Analytics.tsx
mg2239: probably can do in the accessibility branch

mg2239:
good catch with the absolute position lol, i forgot to fix that

lsizemore8 (approved):
Approving because the main issues are fixed. However, the edit button still isn't technically fixed with responsive design. When you adjust the screen width, it starts out fine. However, if you make it too small and then scroll to the left to see the rest of the table, you get the same issue as before (see screenshot). Given web may not need to be that responsive this is a non-issue, hence the approval, but it's something to keep in mind: 
image

mg2239:
hm thats weird because it works fine on my end, not sure why that would be
image

PR #191, opened 2021-04-20
Set up rider web routes

mg2239:
### Summary 

This pull request is the first step to implementing rider web. A route /rider for rider pages has been added, and the Sidebar component has been updated to display the rider web pages. To make the route distinction more clear, /dashboard has been renamed to /admin.

### Test Plan 

image

### Notes 

Right now there's no way for Riders to log into the platform, only admins. This should be addressed in a separate PR.

### Breaking Changes  


helenxuyang (approved):
LGTM, looks like an organized and convenient way to use the site for both admins and riders!

PR #192, opened 2021-04-22
Implemented backend and frontend for student usage

aaronkang7:
### Summary 

This pull request implements backend endpoints and frontend elements for usage element for the students' table. 

Screen Shot 2021-04-21 at 9 12 44 PM

### Breaking Changes  

### DIFF
--- a/frontend/src/components/UserTables/StudentsTable.tsx
+++ b/frontend/src/components/UserTables/StudentsTable.tsx
@@ -14,10 +14,10 @@ const StudentsTable = () => {
   const colSizes = [1, 0.75, 0.75, 1.25, 1];
   const headers = ['Name / NetId', 'Number', 'Address', 'Usage', 'Disability'];
   const [usage, setUsage] = useState({ studentRides: 0, noShowCount: 0 });
-  const getUsageData = (id: string) => {
-    fetch(`/api/riders/${id}/usage`, withDefaults())
-      .then((res) => (res.json()))
-      .then((data) => setUsage(data));
+  const getUsageData = () => {
+    fetch('/api/riders/usage', withDefaults())
+      .then((res) => (console.log(res.json())));
+    // .then((data) => setUsage(data));
     return {
       data:
         <div className={styles.usage}>
@@ -61,7 +61,7 @@ const StudentsTable = () => {
         const disability = accessibility.join(', ');
         const phone = fmtPhone(phoneNumber);
         const shortAddress = address.split(',')[0];
-        const usageData = getUsageData(id);
+        const usageData = getUsageData();
         const riderData = {
           firstName,
           lastName,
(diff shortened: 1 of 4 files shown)


server/router/rider.ts
```diff
@@ -80,6 +80,26 @@ router.get('/:id/currentride', validateUser('Rider'), (req, res) => {
...
+    const conditionNoShow = isRider.group(isNoshow);
+    const isCompleted = new Condition('status').eq(Status.COMPLETED);
+    const conditionRides = isRider.group(isCompleted);
+    db.scan(res, Ride, conditionNoShow, (data: RideType[]) => {
+      noShowCount = data.length;
+      db.scan(res, Ride, conditionRides, (data_: RiderType[]) => {
+        studentRides = data_.length;
+        res.send({ studentRides, noShowCount });
+      });
+    });
```
mg2239, line 99: to reduce database reads, you could use the isRider condition, then do two .filter() calls, one for no show, one for completed, on the data sent from the callback

frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -1,13 +1,38 @@
...
+        <span>
+          <span className={styles.usageContainer}>
+            <span className={cn(styles.ridesCount, styles.usageTag)}></span>
+            {usage.studentRides} Rides
+          </span>
+          <span className={styles.usageContainer}>
+            <span className={cn(styles.noShow, styles.usageTag)}></span>
+            {usage.noShowCount} No Shows
+          </span>
+        </span>,
```
mg2239, line 32: can you make them on the same line like in the designs? should be just a flexbox w/ some positioning

mg2239 (changes requested):
nice work! good job figuring out the server stuff, just a couple small changes

frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -1,13 +1,38 @@
...
+            <span className={cn(styles.ridesCount, styles.usageTag)}></span>
+            {usage.studentRides} Rides
+          </span>
+          <span className={styles.usageContainer}>
+            <span className={cn(styles.noShow, styles.usageTag)}></span>
+            {usage.noShowCount} No Shows
+          </span>
+        </div>,
+    };
+  };
```
mg2239, line 34: just found a bug: usage is the state for the whole table, not just an individual rider. so every rider has the same usage at the moment. we'd either need to make a request each per user (which would be expensive) and figure out a way to keep track of that in state, or create another endpoint to get the usage of all riders in the table. if you have time id like you to try the second option, if not i can try to do it within the next couple days

mg2239 (changes requested):

frontend/src/components/UserTables/StudentsTable.tsx
aaronkang7: i may have some time to work on it this weekend but I don't think I'll be able to finish it by the work session :(

mg2239 (approved):
nice work!

PR #193, opened 2021-04-22
Assign driver

lsizemore8:
### Summary 

I added functionality to the assign driver button in the unscheduled rides. This PR connects this button to the backend, and after clicking, will update the button to show the driver's first initial as the designs show. After a refresh, that ride will be removed from the unscheduled rides and added to the schedule. 

- [x] implemented backend connection to assign driver. 
- [ ] fixed Y

### Test Plan 

To test, either find an unscheduled ride or make a new one in postman, and assign a driver. Here are two screenshots showing after a driver is assigned but the page has not been refreshed & what happens after the page is refreshed. 
assign_driver_apr_22
assign_driver_apr_22_after_refresh

### Notes 

In this moment, the driver background color is always red. I am confirming with designs what that color should be right now (in the designs they always seem to be red). I will either update this PR or make a new one once I get more clarification.

Additionally, the text may be slightly off center in the circle (I don't believe it is but it's always hard to tell). The best way to fix that may be to make the circle a tad smaller than what the design is so the math to get the center of the circle is simple and accurate. 

Finally, the font color in the driver initial is different than the designs. This is on purpose to ensure proper contrast ratio.  

### Breaking Changes  


lsizemore8:
Putting this to WIP because there's an accessibility issue with the dropdown

lsizemore8:
Discussed offline, the accessibility issue will be fixed in a separate PR due to the logic changes required to fix this

mg2239 (approved):

PR #196, opened 2021-04-23
fix incorrect today and tomorrow values on mini cal

lsizemore8:
### Summary 

This pull request fixes an issue with the mini cal. In the mini cal, because it checked if it was "today" or "tomorrow" based on the number in the day only so if today was April 22, going to March 22 or any other month/year with the day being 22 would also make it be "today". 

- [] implemented X
- [x] fixed Mini Cal

### Test Plan 

I tested the calendar by going to a different month with the day being 22 and a different year and ensured that the value was correct. I also tested the "today" and "tomorrow" buttons to ensure correctness. 
Correct Day
fix_mini_cal_apr_22
Different month
fix_mini_cal_apr_22_corr
Different year
fix_mini_cal_apr_22_check_year

### Breaking Changes  

### DIFF
--- a/frontend/src/components/MiniCal/MiniCal.tsx
+++ b/frontend/src/components/MiniCal/MiniCal.tsx
@@ -6,9 +6,14 @@ import './datepicker_override.css';
 import styles from './minical.module.css';
 import { useDate } from '../../context/date';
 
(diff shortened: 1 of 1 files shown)


mg2239 (approved):
nice catch 👍

PR #149, opened 2021-03-22
Create new table component

mg2239:
### Summary 

This pull request aligns table styles better with designs and makes creating tables easier by creating them through CSS grid instead of HTML tables.

### Test Plan 

image
image
image
image

### Notes 

The Students table has been reimplemented in this PR to demo the features. Future PRs will be updating the tables one at a time instead.

Tag colors are currently placeholder, as the colors in designs are not very clear.

Because CSS subgrids are not supported on most browsers yet, the nested grid implementation I used is a bit hacky, but it should work for most cases.

### Breaking Changes  

<!-- - Database schema change (anything that changes DynamoDB document structure)
<!-- - Other change that could cause problems (Detailed in notes)

PR #200, opened 2021-04-24
Refactor RidesTable to use new table components

pihucode:
### Summary
This PR refactors `RidesTable` to use the new table components in #149 

### Test Plan 
Remove the `date=${today}` query from the fetch endpoint in `ScheduledTable` and `UnscheduledTable` components to see all rides.

image

image

### Notes 
Modified `ScheduledTable` props to support both rider and driver information since `ExportPreview` also uses `ScheduledTable`.

Old props:
```
type ScheduledTableProp = {
  driverId: string;
  driverName: string;
};
```

Generalized props:
```
type ScheduledTableProp = {
  query: string; // either 'rider' or 'driver'
  id: string;
  name: string;
};
```

### Breaking Changes 
Added `width: '100%'` to `TableComponent` to fix the row width from being too small. This does not affect any of the existing tables.

image


mg2239 (changes requested):
sorry about the poor documentation for the table components, there's a prop called `groupStart` that implements the separated row sections, so you didn't need to add it in yourself. 

i fixed the unscheduled table component, but noticed theres no scheduled table. could you add that in too, with the driver sections? you shouldn't need to do anything special for it, all table components are just divs.

mg2239 (commented):
looks good, just waiting to get confirmation from PMs/designers whether the rides should be grouped by rider instead of driver

mg2239 (changes requested):
yep should be organized by driver, so just swap so the headers are drivers and columns are rider

pihucode:
> yep should be organized by driver, so just swap so the headers are drivers and columns are rider

Done!

mg2239 (approved):

PR #201, opened 2021-04-27
Get stats endpoint

simranp723:
### Summary 

This PR adds an endpoint that returns ride statistics for the days between (and including) `from` and `to`. It queries the Stats table for the data for these dates, or computes the data from the Rides table if there is no Stats entry for a date.

This PR also removes the "dailyTotal" field from the Stats schema.

### Test Plan 

Call `GET / api/stats/:from/:to`

### Notes 

Ride cancellation is currently not implemented, so this returns 0 for dayCancel and nightCancel

### DIFF
--- a/server/router/stats.ts
+++ b/server/router/stats.ts
@@ -1,0 +1,129 @@
+import express, { Response } from 'express';
+import moment from 'moment-timezone';
+import { Condition } from 'dynamoose/dist/Condition';
+import { AnyDocument } from 'dynamoose/dist/Document';
+import { Stats } from '../models/stats';
+import { Ride } from '../models/ride';
+import * as db from './common';
+import { validateUser } from '../util';
+import { Driver, DriverType } from '../models/driver';
+
+const router = express.Router();
+const tableName = 'Stats';
+
+router.get('/:from/:to', (req, res) => {
+  const { params: { from, to } } = req;
+
+  let date = from;
+  while (date <= to) {
+    const year = moment.tz(date, 'YYYY-MM-DD', 'America/New_York').format('YYYY');
+    const monthDay = moment.tz(date, 'YYYY-MM-DD', 'America/New_York').format('MMDD');
+    const condition = new Condition()
+      .where('year')
+      .eq(year)
+      .and()
+      .where('monthDay')
+      .eq(monthDay);
+
+    const dateMoment = moment.tz(date, 'America/New_York');
+    // day = 12am to 5:00pm
+    const dayStart = dateMoment.toISOString();
+    const dayEnd = dateMoment.add(17, 'hours').toISOString();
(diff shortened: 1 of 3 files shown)


server/router/stats.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Stats } from '../models/stats';
+import { Ride } from '../models/ride';
+import * as db from './common';
+import { validateUser } from '../util';
+import { Driver, DriverType } from '../models/driver';
+
+const router = express.Router();
+const tableName = 'Stats';
+
+router.get('/:from/:to', validateUser('User'), async (req, res) => {
```
mg2239, line 13: from and to should be queries instead of part of the URL

server/router/stats.ts
```diff
@@ -0,0 +1,155 @@
...
+  const { params: { from, to } } = req;
+
+  let date = from;
+  const dates = [];
+  while (date <= to) {
```
mg2239, line 18: `to` may not always be provided, which in that case `dates` should just be `[from]`

server/router/stats.ts
```diff
@@ -0,0 +1,155 @@
...
+  dayStart: string,
+  dayEnd: string,
+  nightStart: string,
+  nightEnd: string,
+  year: string,
+  monthDay: string,
+  date: string,
+  to: string,
+) {
+  db.scan(res, Stats, condition, async (data: Document[]) => {
```
mg2239, line 58: since year and monthDay are keys, you dont need to do a scan

server/router/stats.ts
```diff
@@ -0,0 +1,155 @@
...
+        const dayCountStat = dataDay.length;
+
+        db.scan(res, Ride, conditionRidesDayNoShow, ((dataDayNoShow: Document[]) => {
+          const dayNoShowStat = dataDayNoShow.length;
+
+          db.scan(res, Ride, conditionRidesNight, ((dataNight: Document[]) => {
+            const nightCountStat = dataNight.length;
+
+            db.scan(res, Ride, conditionRidesNightNoShow, ((dataNightNoShow: Document[]) => {
+              const nightNoShowStat = dataNightNoShow.length;
```
mg2239, line 111: to reduce database reads, could you just get all rides w/ stats not generated between `to` and `from`, then do all the filtering with that?

mg2239 (changes requested):
havent tested the endpoint yet but just a few initial thoughts

server/router/stats.ts
simranp723: I think I addressed all comments except this, bc I think in order to do this a GSI has to first be created that combines year and monthDay

mg2239 (approved):
made some fixes since the endpoint wasnt returning the right data. but really nice work!

PR #204, opened 2021-04-28
Date filters on the analytics page

Becky-Hu-82:
### Summary 

This pull request implements the date filters on the analytics page.

- implemented a month picker and a year picker
- implemented a date range picker
- added the analytics overview component and its functionality
image
image


mg2239 (changes requested):
can you better align these to the figma styles?

Becky-Hu-82:
> I was wondering if you can use the icons (i.e. the calendar icon) provided in the Figma to customize the date-picker? (I can't run this on my end at the moment because of some error). But other than that it looks great to me!

That's a good point! Just changed the icon!

mg2239 (changes requested):
image
im getting a dependency error with react-dropdown-date and our version of react. can you also try reinstalling dependencies and see? if its also happening on your end, would it be possible to use native html inputs instead?

frontend/src/components/AnalyticsTable/AnalyticsTable.tsx
```diff
@@ -73,6 +77,94 @@ const Table = ({ type }: TableProps) => {
...
+      // should never get here
+      return [{
+        icon: dayRide,
+        alt: 'day',
+        stats: 0,
+        description: 'day rides',
+      }]
+    }
+  }
+
```
mg2239, line 167: this logic should be in the analytics overview component, not in the table component

frontend/src/components/AnalyticsTable/AnalyticsTable.tsx
```diff
@@ -208,6 +300,9 @@ const Table = ({ type }: TableProps) => {
   }, [type, drivers.length]);
 
   return (
+    <>
+    <DateFilter/>
+    <AnalyticsOverview stats={generateStatsArr()} />
```
mg2239, line 305: move these to the analytics page in Admin. when connecting to backend, the analytics page will handle getting stats and passing the table data to all components, including the table and analytics overview

mg2239 (changes requested):
overall looks good, but just a few changes to better separate the component logic

mg2239 (approved):
swapped out inputs to just be native, other changes require a lot of refactoring so that can be handled in a separate PR

PR #205, opened 2021-04-28
Put stats

simranp723:
### Summary 

This PR implements an endpoint to edit Stats fields for a series of dates. It takes in a map of the dates to be edited with the fields to be updated, and it iterates through each date and updates the Stats entry with the given fields.

### Test Plan 

Call `PUT /api/stats` with a request body in the form { dates: { [date in MM/DD/YYYY format]: { stats fields to be changed },}
- Ex: {"04/05/2021":{"dayCount":0,"nightCount":0},"04/06/2021":{"dayCount":0,"nightCount":0}}

### Notes 
- Date is in MM/DD/YYYY format


server/router/stats.ts
```diff
@@ -0,0 +1,27 @@
...
+  const datesObject = JSON.parse(dates);
+
+  for (const date in datesObject) {
+    if (date) {
+      const year = moment.tz(date as string, 'MM/DD/YYYY', 'America/New_York').format('YYYY');
+      const monthDay = moment.tz(date as string, 'MM/DD/YYYY', 'America/New_York').format('MMDD');
+      const operation = { $SET: datesObject[date] };
+      const key = { year, monthDay };
+
+      db.update(res, Stats, key, operation, tableName);
```
mg2239, line 22: might be better to do a batch update?

mg2239 (changes requested):

server/router/stats.ts
mg2239, line 22: make sure this only sends a response after all updates are made, because rn it send a response after the first entry is updated. feel free to just use the Stats model if the common functions are too restrictive

mg2239 (changes requested):

server/router/stats.ts
simranp723: Ok fixed this using same method as get stats

mg2239 (approved):

PR #207, opened 2021-04-29
Accessibility fixes part 1

lsizemore8:
### Summary 

This is the first PR towards making the admin portal AA accessible. This involved a variety of minor fixes including: 
- Adding focus CSS to all interactable elements
- Fixing html to be interactable via keyboard 
- Adding a skip link (a link that skips repeatable content) 
- Fixed color contrast issues
- Added useful alt text

- [] implemented X
- [x] fixed numerous accessibility bugs

### Test Plan 

Tested every page using the SiteImprove & screen reader. With siteImprove, checked that remaining issues were either guidance (ie "check that x happens") or AAA level issues. Also attempted to use the keyboard to navigate through the website. 

### Notes 

This is only part one of all accessibility fixes. Namely, the following fixes still need to be implemented/fixed:
1. Semantic HTML has started to be added in but it is not complete
2. On the user detail pages, it will always default to going to main content. This is a temporary fix to prevent the page from crashing if the "skip to main content' link is clicked
3. the dropdown on the assign driver modal is not accessible. This will be implemented after the ride table new components to prevent merge conflicts and coding issues
4. The most recent PRs will likely have to be checked over for accessibility purposes
5. A footer required by Cornell needs to be added. That will be in a separate PR to prevent this one from growing any larger.  
6. The alt text and the link alt text are identical for the side bar. I'm still determining if this is an issue or a nonissue. 
7. The rider detail edit button is not accessible. Will add that in part two once edit rider is in. 

### Breaking Changes  

If you use a useLocation/useState in the future, for now ensure the link goes to "pathname#main" as when clicking a "skip to main content" it will add a main link at the end, which may break the use of useState in a link. Hopefully will find a better fix soon.


helenxuyang:
NVDA testing comments:

Home page calendar:
- seems to only navigate between Sun, Wed, and Sat?
- buttons for switching months select a date from the next month and close the modal
- can use tab/arrows to get out of the calendar modal which shouldn't be allowed

Home page:
- Unscheduled Rides should probably be a header?

Nav bar:
- reading kind of strangely, like "students graphic student link"
- right after switching to a different page reads "blank" 
- and then if you hit tab immediately after, doesn't read the "skip to main content" button

Add ride modal: 
- driver selection reads all the drivers' names and then the first radio button
- only the first radio button is accessible when using tab
- arrow keys do switch between radio buttons, but don't read the selected name and also change the selection which I don't think is supposed to happen

Add/edit employee modal:
- edit fields read strangely, like "name edit has autocomplete name blank"
- "Add an Employee" and Edit Profile" should probably be headings
- repeat days don't read when selected/unselected like a checkbox should
- same issue with role radio buttons as the add ride modal

Employees:
- each item on the card reads as "link ___"

Individual employee page:
- might want name to be a heading?
- phone number and needs could probably be read out better like "phone number" instead of "graphic phone" and removing the bullet point in the semantic label for availabilities
- columns read the name twice, like "name column 1 name" "date column 2 date"

Students:
- each item in the row reads as "link ___"

Individual student page:
- might want name to be a heading?
- phone number and needs could probably be read out better instead of "graphic phone" 
- columns read the name twice, like "name column 1 name" "date column 2 date"

Analytics:
- tabbing from Rider Data to Driver Data works, but arrow keys skip from Rider Data to notification button
- when editing, right now it reads like "total 29 row 2 daily ride count column 3, total spin button editable selected 29" so could try to customize so that it just reads column header and date?

lsizemore8:
Current plans based on comments: 
1. Fixing the 5 modals (calendar, add ride, add rider, assign driver, add employee) will be done in a separate PR
2. With regards to the radio buttons, I believe this close to correct behavior (https://www.w3.org/wiki/RadioButton#:~:text=When%20focus%20moves%20to%20the,radio%20buttons%20in%20the%20group.). 
3. the nav bar "blank" issue is due to the hacky nature of the skip link. Will try to fix that in the second accessibility pr


I'll aim to fix the rest of the issues in here.

mg2239 (approved):
while there are still changes to be made, this looks very good for the first part of fixes 👍

PR #209, opened 2021-04-29
changed region

mly32:
### Summary 

Updating sns configuration to create endpoints.

### Breaking Changes  

Region for sns is us-east-1 instead of us-east-2


tyang98 (approved):
image

PR #210, opened 2021-05-04
connected to db

mly32:
### Summary 

created subscription table for the backend.

### Breaking Changes  


mg2239 (dismissed):

server/router/notification.ts
```diff
@@ -58,99 +57,108 @@ const sendMsg = (sub: Subscription, msg: string) => {
...
-    .catch((err) => {
-      res.status(500).send({ err: 'could not send all messages' });
-    });
 });
 
 // subscribe a user
 router.post('/subscribe', validateUser('User'), (req, res) => {
-  if (badPlatform(req.body.platform)) {
+  const { platform } = req.body;
+  if (platform !== PlatformType.WEB && platform !== PlatformType.ANDROID) {
```
tyang98, line 151: If I try to subscribe on an iOS device, wouldn't this result in an invalid platform error?

tyang98 (approved):
Looks good to me, although I had that one question about the device platforms under the subscribe endpoint.

server/router/notification.ts
mly32: We are using GCM for both android and IOS right now, so the platform type is ANDROID for both, which is a bit confusing.

PR #217, opened 2021-05-09
Fix sign in flicker issue

mg2239:
### Summary 

This pull request fixes the sign in flicker issue by tracking the user type in local storage.

### Notes 

As user type is not sensitive data, it's okay to use local storage. In the future, it would be better to separate the two login buttons to different pages.

### Breaking Changes  

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,8 +1,17938 @@
 {
   "name": "carriage-web",
   "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


lsizemore8 (approved):
This seems to work for me (I've tested going to each of the pages & refreshing them and it seems to not flicker/log me back in after showing the landing page once.

PR #221, opened 2021-05-10
Upcoming and Past Rides Table for Rider Web

pihucode:
### Summary 
This PR implements the upcoming and past rides tables for Rider Web.
- [x] Waiting on designers to approve current design (update: design has been approved)

### Test Plan 
If your rider account does not have rides, replace `id` in `RiderHome` with a hardcoded rider id such as `'e4586b99-5dce-47ac-b27d-0fbb3f8b4389'`

image

image

### Breaking Changes  
The `Ride` type has been changed to include new fields.
```
type Ride = {
  id: string;
  type: string;
  status: string;
  late?: boolean;
  startLocation: Location;
  endLocation: Location;
  startTime: string;
  endTime: string;
  rider: Rider;
  driver?: Driver;
  recurring: boolean;
  recurringDays?: number[];
  endDate?: string;
  deleted?: boolean;
  edits?: string[];
};
```


frontend/src/components/UserTables/RiderRidesTable.tsx
```diff
@@ -0,0 +1,130 @@
...
+            let d = new Date(date),
+              month = '' + (d.getMonth() + 1),
+              day = '' + d.getDate(),
+              year = d.getFullYear();
+            if (month.length < 2) 
+              month = '0' + month;
+            if (day.length < 2) 
+              day = '0' + day;
+            return [month, day, year].join('/');
+          };
```
mg2239, line 55: you can use moment().format() for this

mg2239 (changes requested):
have yet to look through all of this PR, but it seems like no future repeating instances of a ride are being displayed in the scheduled table. we need this in case a rider needs to edit a repeating instance of a ride. 

to create the future instances, we just need the frontend to generate fake ride rows on each `recurringDay` the ride occurs on until `endDate`. all dates in `deleted` are days the ride should not be shown on. helen has implemented this for rider mobile, so feel free to ask her or me if you need any clarification on how to do that

frontend/src/types/index.ts
```diff
@@ -103,4 +103,9 @@ export type Ride = {
   endTime: string;
   rider: Rider;
   driver?: Driver;
+  recurring: boolean;
+  recurringDays?: number[];
+  endDate?: string;
+  deleted?: boolean;
```
mg2239, line 109: oops i think the API docs might not be up to date, we changed `deleted` to be an array of string dates

mg2239 (approved):
updated from master to make sure things worked correctly, and fixed some issues with filtering and sorting and responsive design. overall very nice work! just curious why the design in the figma wasn't used, since it seems doable to me. but rn it seems like it might take too much time to redo, so it should be good for now.

pihucode:
> updated from master to make sure things worked correctly, and fixed some issues with filtering and sorting, responsive design, and accessibility. overall very nice work! just curious why the design in the figma wasn't used, since it seems doable to me. but rn it seems like it might take too much time to redo, so it should be good for now.

Thank you for the fixes! I talked with the designers and apparently the design on Figma is outdated so the implemented design is the desired one (includes a time column, groups rides by date, no background color for the repeating dates column).

PR #222, opened 2021-05-10
working sending notifications

mly32:
### Summary 

Notifications are now automatically sent out when a ride changes.

### Test Plan 

Change proxy to the local server, run server and client and edit a ride. It will cause a notification.

(types for subscriptions have changed so call the delete subscriptions function if something goes wrong)

### Notes 

As of right now, notifications are only sent when a ride is edited.

The content of the notification is extremely simple for testing.

Here userType is the user who made the edit and rider is the current ride's rider, driver is the current ride's driver.

The notification is sent to all admin.

If an admin made the edit, the driver and rider get notified as well. If the driver made the edit, the rider also gets one.  If the rider made the edit, the driver gets one if there is a driver assigned.

### Breaking Changes  

Subscriptions now require a user ID.

### DIFF
--- a/server/router/ride.ts
+++ b/server/router/ride.ts
@@ -9,6 +9,8 @@ import { Tag } from '../models/location';
 import { createKeys, validateUser } from '../util';
 import { DriverType } from '../models/driver';
 import { RiderType } from '../models/rider';
+import { UserType } from '../models/subscription';
+import { sendToUsers } from '../util/notification';
 
 const router = express.Router();
 const tableName = 'Rides';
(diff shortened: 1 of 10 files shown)


server/router/ride.ts
```diff
@@ -170,14 +181,50 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+    }
+    if (userType === UserType.DRIVER) {
+      sendToUsers(info, UserType.RIDER, riderId);
+    }
+    res.send(ride);
+  });
 });
 
 // Create edit instances and update a repeating ride's edits field
 router.put('/:id/edits', validateUser('User'), (req, res) => {
```
mg2239, line 217: are you planning on adding repeating ride edit notifications in a separate PR, or can you do it in this one?

mg2239 (approved):
works well, nice job! just a consideration for the future: as the notification code becomes more complex, might be good to add some more comments/documentation just to make it easier to understand and follow

server/router/ride.ts
mly32: Yeah, I'm not entirely sure of the full functionality of the repeating rides endpoint, so in a different PR. At that point, I would probably be abstracting this to another function defined in the notifications utility file.

PR #225, opened 2021-05-12
Fix get stats endpoint

mg2239:
### Summary 

This pull request fixes the get stats endpoint. The function call that generates and sends the response got removed, and the `from`/`to` match checks didn't account for `from` or `to` being undefined.

### Breaking Changes  


helenxuyang (approved):
LGTM, logic makes sense

PR #227, opened 2021-05-15
Cancel ride

zack-ashen:
### Summary

Added ability to cancel rides and trash button next to close modal button on calendar modal. 

This pull request is the first step towards implementing feature canceling rides.

- [x] Cancel normal rides
- [x] Cancel recurring rides
- [x] Add trash icon to schedule modal

### Test Plan

Canceled both recurring and normal rides to confirm they delete and update state to reflect those changes.


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -163,6 +168,35 @@ Rider: ${ride.rider.firstName} ${ride.rider.lastName}`,
...
+          withDefaults({
+            method: 'DELETE',
+            headers: { 'Content-Type': 'application/json' },
+          }),
+        )
+        .then(() => getRides())
+        .then(closeModal);
+      }
+    })
+  }
```
mg2239, line 198: couldnt you pass the ride object instead of refetching the ride?

mg2239 (changes requested):

frontend/src/components/Schedule/Schedule.tsx
zack-ashen: I had to refetch the ride because type Ride doesn't have the recurring field. I could add it but I figured that would cause errors elsewhere.

mg2239: ah i see, if you can add it that would be good, but if theres errors you can cast it to `any` just as a temp fix for the time being

mg2239 (approved):
moved filtering cancelled rides out to the backend, but overall looks good!

PR #229, opened 2021-05-17
fix broken changes from previous pr

zack-ashen:
### Summary 

Fixed problems from previous pr.

- [x] Added tags
- [x] Added back edit employee
- [x] Added hours driving statistics


mg2239 (approved):
fixed a couple more issues, namely fetching user information when trying to access a user link directly and adding the edit an employee back (but didn't notice you made that fix at the same time). otherwise looks good to me

PR #231, opened 2021-05-17
Various fixes and improvements

mg2239:
### Summary 

This pull request implements a lot of small changes and improvements to the codebase, including:

- [x] Logout on sidebar no longer overlaps with other icons when screen height is too small
(however it is now no longer accessible past a certain screen height, so a responsive sidebar is needed)
- [x] Collapsible component styles updated
- [x] Unscheduled and Scheduled ride tables do not display anything if there are no rides in table
- [x] Reduce database reads in Scheduled table
- [x] Ride status modal uses new Tag component
- [x] Locations table sorts alphabetically correctly
- [x] Formatting fixes for Home page, Tag component
- [x] Driver stats endpoint now only generates stats based on completed rides within past week
- [x] Deleted old/unused components which should reduce conflicts when updating types
- [x] Update all types to match server
- [x] Employee startDate prefilled when editing employee

### Test Plan 

image
image
image
image
image

### Breaking Changes  


frontend/src/pages/Admin/Locations.tsx
```diff
@@ -15,6 +15,11 @@ const Locations = () => {
       const locationsData = await fetch('/api/locations', withDefaults())
         .then((res) => res.json())
         .then((data) => data.data);
+      locationsData.sort((a: Location, b: Location) => {
+        if (a.name < b.name) { return -1; }
+        if (a.name > b.name) { return 1; }
+        return 0;
```
helenxuyang, line 21: Super minor, but could this be combined into something like `a.name.localeCompare(b.name)` ? (not 100% sure if that's the function)

helenxuyang (approved):
Overall looks great, ran locally to check UI changes and everything works as described. The sidebar only becomes inaccessible when I make the window height very small which seems like an edge case, but maybe it will be made scrollable or something later on. Thank you Matt for the code cleanup and fixes!

frontend/src/pages/Admin/Locations.tsx
mg2239: apparently localeCompare is much slower, but nice catch! i didnt even know that existed until now lol

PR #233, opened 2021-05-23
updated schedule view with correct colors and arrows

mly32:
### Summary 

- Fixed schedule rides so it uses a color map. It is currently based on the driver id (colors may need to be updated if designs have changed since last semester).
- Changed view more state to also use buttons (8hr range).

image
Here, the rides for a driver are given yellow.

### Test Plan 

Add scheduled rides and check colors + time ranges (arrows should be disabled when they are no longer allowed).

### Notes 

On the Figma designs, the rides have colors that are not based on driver id, but I am not sure how those colors are being determined. For now, I think it's fine to go by the driver?

### Breaking Changes  


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -86,45 +101,52 @@ Rider: ${ride.rider.firstName} ${ride.rider.lastName}`,
...
-  const updateRides = (rideId: string, updatedDriver: Driver) => {
-    fetch(
-      `/api/rides/${rideId}`,
-      withDefaults({
-        method: 'PUT',
-        headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({ driver: updatedDriver }),
-      }),
-    );
-  };
```
mg2239, line 98: we dont actually have an easy way to reassign drivers other than drag and drop, so we'll need to include it for now

mg2239 (changes requested):

mly32:
I attempted to fix many bugs.

I first checked over the lines with eslint disabled and fixed those (added some useCallbacks)

There were many warnings related to [https://stackoverflow.com/questions/58038008/how-to-stop-memory-leak-in-useeffect-hook-react/58038029
](https://stackoverflow.com/questions/58038008/how-to-stop-memory-leak-in-useeffect-hook-react/58038029
) which was fixed with a check for mount before assignment.

There is still a bug for [https://github.com/anthonyjgrove/react-google-login/issues/311](https://github.com/anthonyjgrove/react-google-login/issues/311) that I attempted to fix by checking if the profile path was defined.

There are some bugs when loading in the rider pages for some reason that I could not resolve.

mg2239 (approved):
looks really nice, thanks for getting to all the bugs! i made some other style changes and had to revert one of the useCallback fixes you did, since it resulted in editing an employee breaking. its probably doable with useCallback, but i think we can leave it for later.

PR #245, opened 2021-06-27
Added submit button and error text for analytics date selection

helenxuyang:
### Summary 

- Removed year and month filters because these are included in the start and end dates
- Added apply dates button (no longer automatically updates table on date change to help with invalid dates)
- Added flex CSS to help with smaller screens/windows

### Test Plan 
image
image
image
image

Confirm that:
- Default start and end dates are the current date
- Start and end date options go up to and include current date
- Selecting start date after end date shows error message
- Selecting end date before start date shows error message
- Selecting valid dates clears error message
- Selecting same date for start and end does NOT show error message
- Table does not update until apply button is clicked
- Table does not update if apply button is clicked when there are errors
- Table updates correctly according to dates selected


mg2239 (approved):
refactored a bit to remove the need for useEffect, really nice work!

PR #251, opened 2021-09-13
Fix build + add CI check to prevent future errors

mg2239:
### Summary 

This pull request fixes our CI to make sure both frontend and backend can be built before a PR can be merged. This will prevent future errors when a PR is good to be merged but unexpectedly breaks staging/prod. In this case, `react-scripts` was out of date, and there were some devDependencies that should be in dependencies, so this PR also fixes that.

### Breaking Changes  


aaronkang7 (approved):
Looks good!

PR #262, opened 2021-10-07
Commented out the trash/delete button in UserDetail

20austinw:


20austinw:
Removed trash/delete button

PR #267, opened 2021-10-14
Updated User ID checking

justinkang01:
### Summary 

This pull request is to check the userID of a rider and/or driver to only allow the verified user to edit their own data.

### Breaking Changes  

### DIFF
--- a/server/router/driver.ts
+++ b/server/router/driver.ts
@@ -160,7 +160,10 @@ router.put('/:id', validateUser('Driver'), (req, res) => {
     params: { id },
     body,
   } = req;
-  if (res.locals.userType == UserType.ADMIN || id === res.locals.user.id) {
+  if (
+    res.locals.user.userType === UserType.ADMIN ||
+    id === res.locals.user.id
+  ) {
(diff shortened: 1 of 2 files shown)


server/router/driver.ts
```diff
@@ -159,7 +159,11 @@ router.put('/:id', validateUser('Driver'), (req, res) => {
     params: { id },
     body,
   } = req;
-  db.update(res, Driver, { id }, body, tableName);
+  if (id === res.locals.user.id) {
```
mg2239, line 162: we want admins to be able to edit all users, so allow edits if `userType === UserType.ADMIN` too

server/router/rider.ts
```diff
@@ -171,7 +171,11 @@ router.put('/:id', validateUser('Rider'), (req, res) => {
     params: { id },
     body,
   } = req;
-  db.update(res, Rider, { id }, body, tableName);
+  if (id === res.locals.user.id) {
```
mg2239, line 174: same here

mg2239 (changes requested):

PR #270, opened 2021-10-20
Added a toast when editing an employee.

tuckerstanley:
### Summary 

This pull request adds a toast when editing an employee. Previously, there was no toast, so I implemented one. 

### Breaking Changes  

### DIFF
--- a/frontend/src/components/EmployeeModal/EmployeeModal.tsx
+++ b/frontend/src/components/EmployeeModal/EmployeeModal.tsx
@@ -142,6 +142,18 @@ const EmployeeModal = ({ existingEmployee }: EmployeeModalProps) => {
     refresh: () => Promise<void>,
     table: string
   ) => {
+   fetch(
+      endpoint,
+      withDefaults({
+        method: 'PUT',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(employeeData),
+      })
+    ).then(() => {
+      refresh();
+      setToast(true);
+    });
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -142,6 +142,18 @@ const EmployeeModal = ({ existingEmployee }: EmployeeModalProps) => {
     refresh: () => Promise<void>,
     table: string
   ) => {
+   fetch(
+      endpoint,
+      withDefaults({
+        method: 'PUT',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(employeeData),
+      })
```
mg2239, line 151: the fetch request is already being handled in `updatedEmployee`, dont need to do it twice. you can do something like 
```
then((res) => {
  refresh();
  setToast(true)
  return res.json()
})
```

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -157,7 +169,7 @@ const EmployeeModal = ({ existingEmployee }: EmployeeModalProps) => {
   const onSubmit = async (data: ObjectType) => {
     const { firstName, lastName, email, phoneNumber, startDate, availability } =
       data;
-
+    console.log(existingEmployee)
```
mg2239, line 172: remove console.log

PR #271, opened 2021-10-21
Met236/date format update

michaeltorku:
Added a function date format to replace the implementation to moment().format() at several points in the code.

### DIFF
--- a/frontend/src/util/index.tsx
+++ b/frontend/src/util/index.tsx
@@ -1,18 +1,10 @@
 import moment from 'moment';
 import React from 'react';
 
-const format_date = (
-  date?: string | Date,
-  default_format?: string,
-  format?: string
-) => {
+export const format_date = (date?: string | Date, format?: string) => {
   const fmt = format ? format : 'YYYY-MM-DD';
   if (date == undefined) {
     return moment().format(fmt);
-  } else if (default_format == undefined) {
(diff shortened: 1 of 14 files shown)


frontend/src/util/index.tsx
```diff
@@ -0,0 +1,18 @@
+import moment from 'moment';
+import React from 'react';
+
+const format_date = (
```
mg2239, line 4: i think export const would work better here, since we might have other functions in this file

frontend/src/util/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+  date?: string | Date,
+  default_format?: string,
+  format?: string
```
mg2239, line 7: you can actually do `format = 'YYYY-MM-DD'` here instead to avoid creating a new variable. this will set format to 'YYYY-MM-DD' if nothing is provided

frontend/src/util/index.tsx
```diff
@@ -0,0 +1,18 @@
+import moment from 'moment';
+import React from 'react';
+
+const format_date = (
+  date?: string | Date,
+  default_format?: string,
```
mg2239, line 6: is default_format ever used in the codebase? if not we dont need to account for it

mg2239 (approved):
lgtm!

PR #285, opened 2021-11-20
Set default timezone for moment

fengyuli2002:
### Summary 

Set default timezone for moment using moment.tz.setDefault().

All previous usages of moment.tz() are replaced with moment().

### Test Plan 

I switched my own laptop's local timezone to London and console logged the result of moment().format(). It turns out to be converted to EST, which is ideal.

### Notes 

It seems that I have to set the default timezone in app.ts before importing other files (riders, stats, etc) so that the default timezone works in those files as well.

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -6,7 +6,7 @@ import cors from 'cors';
 
 // Set default timezone for moment
 import moment from 'moment-timezone';
-moment.tz.setDefault("America/New_York");
+moment.tz.setDefault('America/New_York');
 
 import config from './config';
 import rider from './router/rider';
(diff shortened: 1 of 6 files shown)


fengyuli2002:
How to run prettier? Is it something about code style?
I installed ESlint, not sure if that's relevant.

mg2239:
> How to run prettier? Is it something about code style? I installed ESlint, not sure if that's relevant.

to run prettier, run `npm run format` in carriage-web, and to run eslint run `npm run lint` in carriage-web

fengyuli2002:
> 

Okay I've done that. Also,

Typically who should we refer to for code review?

tuckerstanley (approved):
The code looks good! It seems as if you were just changing all usages of moment.tz() with moment(). Also good job with the testing and changing your timezone. To answer your question, usually you don't need to refer someone in specific, one of the devs or Matt will just look at it in a timely manner. Code seems ready to merge!

PR #289, opened 2021-12-03
jk2375/non-empty-fileBuffer

justinkang01:
### Summary 

Disallow the fileBuffer to be an empty string. Sends a bad request error if fileBuffer happens to be an empty string.

- [x] Implement string checking for fileBuffer

### Breaking Changes  

### DIFF
--- a/server/router/upload.ts
+++ b/server/router/upload.ts
@@ -18,7 +18,9 @@ router.post('/', validateUser('User'), (req, res) => {
     body: { id, tableName, fileBuffer },
   } = req;
   const validTables = ['Riders', 'Drivers', 'Admins'];
-
+  if (fileBuffer === '') {
+    res.status(400).send({ err: 'Invalid file name: empty string' });
+  }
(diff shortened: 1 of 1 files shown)


20austinw (approved):
Clean code! Looks good to merge.

server/router/upload.ts
```diff
@@ -18,7 +18,9 @@ router.post('/', validateUser('User'), (req, res) => {
     body: { id, tableName, fileBuffer },
   } = req;
   const validTables = ['Riders', 'Drivers', 'Admins'];
-
+  if (fileBuffer === '') {
+    res.status(400).send({ err: 'Invalid file name: empty string' });
+  }
```
mg2239, line 23: you actually need an `else` here wrapping the rest of the function, or else the rest of the code will continue to execute

mg2239 (changes requested):

mg2239 (approved):

PR #291, opened 2022-01-03
fl334/available drivers

fengyuli2002:
### Get available drivers

- [x] implemented `/api/drivers/available`

When given a date=YYYY-MM-DD, startTime=hh:mm, and endTime=hh:mm query, this endpoint should return a list of drivers that are available during this date and time range, based on a driver's availability.

### Test Plan 

I called this endpoint on Postman and it runs successfully.
It is able to correctly filter out unavailable drivers.

### Notes 

1. This endpoint assumes there will not be drivers during weekends.
2. An empty list is returned when there is no drivers available.
3. Another possible output is `{err : 'startTime must precede endTime'}`

### DIFF
--- a/server/router/driver.ts
+++ b/server/router/driver.ts
@@ -46,10 +46,7 @@ router.get('/available', validateUser('User'), (req, res) => {
         return false;
       }
 
-      return (
-        availEnd >= reqEndTime &&
-        availStart <= reqStartTime
-      );
+      return availEnd >= reqEndTime && availStart <= reqStartTime;
     });
 
     res.send({ data: drivers });
--- a/server/router/driver.ts
+++ b/server/router/driver.ts
@@ -14,6 +14,7 @@ const tableName = 'Drivers';
 
 // Get a driver by id in Drivers table
 router.get('/:id', validateUser('User'), (req, res) => {
+    console.log("triggered wrong")
   const {
     params: { id },
   } = req;
@@ -45,7 +46,8 @@ router.get('/:id/profile', validateUser('User'), (req, res) => {
 });
 
 // Get all available drivers at a given time
-router.get('/available', validateUser('Admin'), (req, res) => {
+router.get('/test/available', validateUser('Admin'), (req, res) => {
+  console.log("triggered")
   const { date, reqStartTime, reqEndTime } = req.query;
   const startTime = moment(reqStartTime as string, 'HH:mm');
   const endTime = moment(reqEndTime as string, 'HH:mm');
@@ -73,12 +75,13 @@ router.get('/available', validateUser('Admin'), (req, res) => {
     res.status(400).send({ err: 'startTime must precede endTime' });
   }
 
-  const condition = new Condition()
+  const condition = new Condition(`availability.${getDay()}`).exists()
     .where(`availability.${getDay()}.startTime`)
-    .le(startTime)
+    .le(reqStartTime).and()
     .where(`availability.${getDay()}.endTime`)
-    .ge(endTime);
+    .ge(reqEndTime);
 
+  console.log(`availability.${getDay()}`)
   db.scan(res, Driver, condition);
 });
 
--- a/server/router/driver.ts
+++ b/server/router/driver.ts
@@ -49,15 +49,37 @@ router.get('/available', validateUser('Admin'), (req, res) => {
   const { date, reqStartTime, reqEndTime } = req.query;
   const startTime = moment(reqStartTime as string, 'HH:mm');
   const endTime = moment(reqEndTime as string, 'HH:mm');
-  const day = moment(date as string).day()
 
-  if (startTime > endTime) {
-    res.status(400).send({ err: 'startTime must precede endTime'})
+  function getDay() {
+    switch (moment(date as string).day()) {
+      case 0:
+        return 'Sun';
+      case 1:
+        return 'Mon';
+      case 2:
+        return 'Tue';
+      case 3:
+        return 'Wed';
+      case 4:
+        return 'Thu';
+      case 5:
+        return 'Fri';
+      case 6:
+        return 'Sat';
+    }
   }
 
-  const available = {};
+  if (startTime > endTime) {
+    res.status(400).send({ err: 'startTime must precede endTime' });
+  }
 
-  res.status(200).send(available);
+  const condition = new Condition()
+    .where(`availability.${getDay()}.startTime`)
+    .le(startTime)
+    .where(`availability.${getDay()}.endTime`)
+    .ge(endTime);
+
+  db.scan(res, Driver, condition);
 });
 
 // Get whether a driver is available at a given time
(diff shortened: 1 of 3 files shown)


server/router/driver.ts
```diff
@@ -12,6 +12,52 @@ import { UserType } from '../models/subscription';
...
+  }
+
+  db.getAll(res, Driver, tableName, (doc: DriverType[]) => {
+    const drivers = doc.filter((driver) => {
+      const availStart = (() => {
+        if (reqDay === 2) return driver.availability.Tue?.startTime;
+        if (reqDay === 1) return driver.availability.Mon?.startTime;
+        if (reqDay === 3) return driver.availability.Wed?.startTime;
+        if (reqDay === 4) return driver.availability.Thu?.startTime;
+        if (reqDay === 5) return driver.availability.Fri?.startTime;
```
mg2239, line 36: i think your original approach of getting the day as a string (Mon, Tue, etc.), then finding `driver.availability[day].startTime` or `endTime` was better. doing it like this results in a lot of redundant code

server/router/driver.ts
```diff
@@ -12,6 +12,52 @@ import { UserType } from '../models/subscription';
...
+      if (availStart === undefined || availEnd === undefined) {
+        return false;
+      }
+      return (
+        availEnd >= (reqEndTime as string) &&
+        availStart <= (reqStartTime as string)
+      );
+    });
+
+    res.send(drivers);
```
mg2239, line 57: if the response is an array, the body should be `{ data: drivers }`.

mg2239 (changes requested):
nice job getting this done so quickly! had a few comments below, also if you can remove the `'/:id/:startTime/:endTime'` endpoint that would be great too

server/router/driver.ts
fengyuli2002: Done

server/router/driver.ts
```diff
@@ -12,6 +12,52 @@ import { UserType } from '../models/subscription';
...
+// Get all available drivers at a specific date and time
+router.get('/available', validateUser('User'), (req, res) => {
+  const { date, startTime: reqStartTime, endTime: reqEndTime } = req.query;
+  const reqDay = moment(date as string).day();
+
+  if ((reqStartTime as string) >= (reqEndTime as string)) {
+    res.status(400).send({ err: 'startTime must precede endTime' });
+  }
+
+  db.getAll(res, Driver, tableName, (doc: DriverType[]) => {
```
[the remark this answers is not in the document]
fengyuli2002: No. Dynamoose condition is not able to check if a startTime is greater than another time because not all drivers have startTime on that particular day.

server/router/driver.ts
fengyuli2002: Done

server/router/driver.ts
```diff
@@ -20,11 +64,6 @@ router.get('/:id', validateUser('User'), (req, res) => {
   db.getById(res, Driver, id, tableName);
 });
 
-// Get all drivers
-router.get('/', validateUser('Admin'), (req, res) => {
-  db.getAll(res, Driver, tableName);
-});
-
```
mg2239, line 27: why was this removed? we still use this endpoint on the frontend

server/router/driver.ts
```diff
@@ -12,6 +12,50 @@ import { UserType } from '../models/subscription';
...
+// Get all available drivers at a specific date and time
+router.get('/available', validateUser('User'), (req, res) => {
+  const { date, startTime: reqStartTime, endTime: reqEndTime } = req.query;
+  const numToDay = {
+    1: 'Mon',
+    2: 'Tue',
+    3: 'Web',
+    4: 'Thu',
+    5: 'Fri',
+  };
```
mg2239, line 29: nice change! maybe use an array here instead? theres a little too much typecasting going on for `reqDate`. you can put placeholders (like undefined) for array position 0 and 7

server/router/driver.ts
```diff
@@ -12,6 +12,50 @@ import { UserType } from '../models/subscription';
...
 const tableName = 'Drivers';
 
+// Get all drivers
+router.get('/', validateUser('Admin'), (req, res) => {
+  db.getAll(res, Driver, tableName);
+});
+
+// Get all available drivers at a specific date and time
+router.get('/available', validateUser('User'), (req, res) => {
+  const { date, startTime: reqStartTime, endTime: reqEndTime } = req.query;
```
mg2239, line 22: since you repeatedly use `reqStartTime as string` and `reqEndTime as string` multiple times, you should create separate string variables for `reqStartTime` and `reqEndTime`

server/router/driver.ts
```diff
@@ -12,6 +12,50 @@ import { UserType } from '../models/subscription';
...
+
+  if ((reqStartTime as string) >= (reqEndTime as string)) {
+    res.status(400).send({ err: 'startTime must precede endTime' });
+  }
+
+  db.getAll(res, Driver, tableName, (doc: DriverType[]) => {
+    const drivers = doc.filter((driver) => {
+      const availStart = !reqDate
+        ? reqDate
+        : (driver.availability as any)[reqDate]?.startTime;
```
mg2239, line 40: we want to minimize as many `as any` casts as possible, so make sure `reqDay` is typed correctly. you can do this by adding a type to `numToDay`.

mg2239, line 40: we can simplify this by doing `reqDate && ...`
if `reqDate` is undefined, `availStart` will be undefined.

mg2239 (changes requested):
very close!

server/router/driver.ts
fengyuli2002: Actually it's not removed. Now it's the first endpoint in driver.ts.

server/router/driver.ts
fengyuli2002: Done

server/router/driver.ts
fengyuli2002: Done

server/router/driver.ts
fengyuli2002: That's nice trick!

fengyuli2002: Done, but have to import AvailabilityType.

mg2239: thats fine

server/router/driver.ts
```diff
@@ -4,14 +4,58 @@ import { Condition } from 'dynamoose';
...
+      const availStart = reqDate && driver.availability[reqDate]?.startTime;
+      const availEnd = reqDate && driver.availability[reqDate]?.endTime;
+
+      if (availStart === undefined || availEnd === undefined) {
+        return false;
+      }
+
+      return (
+        availEnd >= (reqEndTime as string) &&
+        availStart <= (reqStartTime as string)
```
mg2239, line 51: one last thing, remove the `as string` casts for reqEndTime and reqStartTime

mg2239 (dismissed):
really nice job! one last change but should be good to merge after that

mg2239 (approved):

PR #293, opened 2022-01-20
Add fields to push notif payload, remove rider notifs when reassigning drivers

mg2239:
### Summary 

This pull request adds some additional information to the push notification payload that is needed for the tap notification interaction. The `id` field makes sure the notification does not display more than once on the Notification page, and the `sentTime` field records the time the notification was created.

Additionally, some changes were made to the data sent when reassigning a driver, to prevent riders from getting notifications if a driver is reassigned.

### Test Plan 

The only visible change is with reassigning drivers, so that's the only part that *needs* to be tested.

The push notification payload can be tested using `server/util/notification.test.ts`, using the id of a rider's ride in the database for `unscheduledRide`, but the only change is additional information, so it shouldn't break anything.

### Breaking Changes  

### DIFF
--- a/frontend/src/components/Modal/AssignDriverModal.tsx
+++ b/frontend/src/components/Modal/AssignDriverModal.tsx
@@ -9,6 +9,7 @@ type AssignModalProps = {
   close: () => void;
   ride: Ride;
   allDrivers: Driver[];
+  reassign: boolean;
 };
 
 type DriverRowProps = {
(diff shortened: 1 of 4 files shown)


michaeltorku (approved):
I like that you replaced reassign as a boolean. So now we have a more safe range of values wrt to it. For the id field (we could use the counting numbers for id to have analytics on the number of notifications we send to the riders) - I've made a note of it for a future pull request. The code is correct and works. Everything looks good!

PR #296, opened 2022-01-30
Suggest available drivers only

fengyuli2002:
### Summary

Previously, the ride scheduling modal lists all drivers as available drivers regardless of their actual availability. This PR fixes this and only allows the drivers who's availability matches the date, pickupTime, and dropoffTime selected by the user. Availability data is based on the backend endpoint `/api/drivers/available` which I implemented during the winter break.

Since displaying available drivers is now an asynchronous function, I let the modal first display "loading" and then transition to either "none" or a list of actually available drivers. The process takes about half-second on my local servers.

### Test Plan 

It looks good on my side :stuck_out_tongue_winking_eye:. Here's some pics:
image
image
image

### Notes 

It seems that the ride is not listed under "scheduled riders" somehow. But I switched back to the master branch and the problem still exists, so it's likely not a problem of this PR.

### DIFF
--- a/frontend/src/components/RideModal/Pages/Driver.tsx
+++ b/frontend/src/components/RideModal/Pages/Driver.tsx
@@ -28,9 +28,7 @@ const DriverPage = ({ onBack, onSubmit, formData }: ModalPageProps) => {
         .then((res) => res.json())
         .then((data) => {
           setAvailableDrivers(
-            [{ id: 'None', firstName: 'None', lastName: 'None' }].concat(
-              data.data
-            )
+            [{ id: 'None', firstName: 'None', lastName: 'None' }, ...data.data]
           );
           setLoaded(true);
         });
--- a/frontend/src/components/RideModal/Pages/Driver.tsx
+++ b/frontend/src/components/RideModal/Pages/Driver.tsx
@@ -13,11 +13,14 @@ const DriverPage = ({ onBack, onSubmit, formData }: ModalPageProps) => {
   });
   const { errors } = formState;
   const { withDefaults } = useReq();
-  const [refreshed, setRefreshed] = useState(false);
+  const [loaded, setLoaded] = useState(false);
+  const [hasDrivers, setHasDrivers] = useState(false);
 
   const { date, pickupTime: startTime, dropoffTime: endTime } = formData!;
   type DriverOption = { id: string; firstName: string; lastName: string };
-  const [availableDrivers, setAvailableDrivers] = useState<DriverOption[]>([]);
+  const [availableDrivers, setAvailableDrivers] = useState<DriverOption[]>([
+    { id: 'None', firstName: 'None', lastName: '' },
+  ]);
 
   useEffect(() => {
     if (startTime && endTime && date) {
@@ -28,7 +31,8 @@ const DriverPage = ({ onBack, onSubmit, formData }: ModalPageProps) => {
         .then((res) => res.json())
         .then((data) => {
           setAvailableDrivers(data.data);
-          setRefreshed(true);
+          setLoaded(true);
+          setHasDrivers(data.data.length >= 1);
         });
     }
   });
@@ -37,8 +41,8 @@ const DriverPage = ({ onBack, onSubmit, formData }: ModalPageProps) => {
     <form onSubmit={handleSubmit(onSubmit)} className={styles.form}>
       <div className={styles.inputContainer}>
         <div className={styles.drivers}>
-          {refreshed ? (
-            availableDrivers.length ? (
+          {loaded ? (
+            hasDrivers ? (
               availableDrivers.map((d) => (
                 <div className={styles.driver} key={d.id}>
                   <Label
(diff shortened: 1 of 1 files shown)


frontend/src/components/RideModal/Pages/Driver.tsx
```diff
@@ -1,42 +1,68 @@
...
-                className={styles.driverRadio}
-                name="driver"
-                type="radio"
-                value={d.id}
-                ref={register({ required: true })}
-              />
-            </div>
-          ))}
+          {refreshed ? (
+            availableDrivers.length ? (
```
mg2239, line 41: maybe create separate booleans for each condition for readability, something like `loading`, `hasNoDrivers`, and `hasDrivers`? then you can separate out the conditional rendering instead of nesting them

frontend/src/components/RideModal/Pages/Driver.tsx
```diff
@@ -1,42 +1,68 @@
...
-  const { drivers } = useEmployees();
+  const { withDefaults } = useReq();
+  const [refreshed, setRefreshed] = useState(false);
+
+  const { date, pickupTime: startTime, dropoffTime: endTime } = formData!;
   type DriverOption = { id: string; firstName: string; lastName: string };
-  const driverOptions: DriverOption[] = [
-    { id: 'None', firstName: 'None', lastName: '' },
-  ].concat(drivers);
+  const [availableDrivers, setAvailableDrivers] = useState<DriverOption[]>([]);
```
mg2239, line 20: we want to preserve the "None" option from the original `driverOptions` variable, since admins might not want to assign a driver yet even if there are ones available

frontend/src/components/RideModal/Pages/Driver.tsx
```diff
@@ -1,42 +1,68 @@
...
+                    value={d.id}
+                    ref={register({ required: true })}
+                  />
+                </div>
+              ))
+            ) : (
+              <p>None</p>
+            )
+          ) : (
+            <p>Loading</p>
```
mg2239, line 64: i think "Loading..." would look a bit better, but thats just a small thing lol

mg2239 (changes requested):
really nice work! i just had a few small comments, appreciate the help with frontend tasks 👍

frontend/src/components/RideModal/Pages/Driver.tsx
fengyuli2002: Okay, I will do that.
Am I understanding this correctly? It seems adding the "None" option will make the card never displays the word "none".

frontend/src/components/RideModal/Pages/Driver.tsx
fengyuli2002: Sure; done.

frontend/src/components/RideModal/Pages/Driver.tsx
fengyuli2002: I just tried breaking them done (instead of nested) but that seems to be not working because displaying drivers / None should happen only if it is `loaded`. 
I did make new booleans and change their names so that it's more readable now.

frontend/src/components/RideModal/Pages/Driver.tsx
mg2239: I think just the none option is fine, so the "none" word isnt needed

frontend/src/components/RideModal/Pages/Driver.tsx
mg2239: the booleans can depend on each other, like `loading = ...`, `hasNoDrivers = loading && ...`, `hasDrivers = loading & ...`

frontend/src/components/RideModal/Pages/Driver.tsx
fengyuli2002: okay

frontend/src/components/RideModal/Pages/Driver.tsx
fengyuli2002: It seems we just need `loaded` now because there's always a None option.

fengyuli2002: I think I've made the change; but GitHub won't let me re-request review because I must make the requested change. Anyway to fix this?

frontend/src/components/RideModal/Pages/Driver.tsx
```diff
@@ -1,42 +1,68 @@
...
+      fetch(
+        `/api/drivers/available?date=${date}&startTime=${startTime}&endTime=${endTime}`,
+        withDefaults()
+      )
+        .then((res) => res.json())
+        .then((data) => {
+          setAvailableDrivers(
+            [{ id: 'None', firstName: 'None', lastName: 'None' }].concat(
+              data.data
+            )
```
mg2239, line 33:
```suggestion
            [{ id: 'None', firstName: 'None', lastName: 'None' }, ...data.data]
```

frontend/src/components/RideModal/Pages/Driver.tsx
```diff
@@ -1,42 +1,68 @@
...
+          );
+          setLoaded(true);
+        });
+    }
+  });
```
mg2239, line 38:
```suggestion
  }, [startTime, endTime, date]);
```

mg2239 (dismissed):
just a couple lines id change, nice work!

fengyuli2002:
Hi Matthew, I committed your change and it seems you need to approve again!

mg2239 (approved):

PR #298, opened 2022-02-25
Rider object reduced tag setting

michaeltorku:
### Summary changing the ride tag
Set the reduced tag for the ride object to be the default and position it to be aligned with line-height of the adjacent text.

### DIFF
--- a/frontend/src/components/TableComponents/TableComponents.tsx
+++ b/frontend/src/components/TableComponents/TableComponents.tsx
@@ -54,7 +54,7 @@ export const Row = ({
   header,
   groupStart,
   className,
-  onClick,
+  onClick
 }: RowProps) => {
   const { width } = useWindowSize();
   const isMobile = Boolean(width && width < 700);
@@ -70,8 +70,8 @@ export const Row = ({
       if (typeof cell === 'string') {
         return <Cell key={idx} data={cell} />;
       }
-      const { data: cData, tag, smallTag } = cell;
-      return <Cell key={idx} data={cData} tag={tag} smallTag={smallTag} />;
+      const { data: cData, tag } = cell;
+      return <Cell key={idx} data={cData} tag={tag} smallTag={true} />;
     });
 
   if (!header && groupStart) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/Tag/tag.module.css
```diff
@@ -1,7 +1,10 @@
 .reducedTag {
   height: 11px;
   width: 11px;
-  float: left;
+  display: inline-block;
+  margin-left: 6px;
+  margin-right: 5px;
```
fengyuli2002, line 6: Is there special considerations on these values?
I see the width is already 11px.

frontend/src/components/TableComponents/TableComponents.tsx
```diff
@@ -70,8 +70,8 @@ export const Row = ({
       if (typeof cell === 'string') {
         return <Cell key={idx} data={cell} />;
       }
-      const { data: cData, tag, smallTag } = cell;
-      return <Cell key={idx} data={cData} tag={tag} smallTag={smallTag} />;
+      const { data: cData, tag } = cell;
+      return <Cell key={idx} data={cData} tag={tag} smallTag={true} />;
```
fengyuli2002, line 74: Is it still possible to set tags as not reduced? Or this functionality is just abandoned?

fengyuli2002 (commented):
I think this is a nice fix!
Just have one question on that reduced tag.
More specifically, is there still someway to manually set tags as not reduced?

michaeltorku:
No. For now - in line with the design guidelines - we're withholding that functionality, but I've made a note of it and we will allow the user to set it later when we do a visual layout settings section in the future

frontend/src/components/TableComponents/TableComponents.tsx
michaeltorku: abandoned for now, see my comment below

frontend/src/components/Tag/tag.module.css
michaeltorku: yes they look good and work well (responsively)

fengyuli2002 (approved):
Thanks for your clarification!

PR #299, opened 2022-02-25
Available drivers only allow drivers not already scheduled for a ride

fengyuli-dev:
### Summary 

Endpoint `/available` now has a new rule. It filters out drivers who's availability matches the requested ride but are already scheduled for a ride for that time interval.
This is done by scanning the database using dynamoose conditions and moment.js formats.
Tested by printing out the rides that are filtered out. Result seems reasonable.

### Breaking Changes  

### DIFF
--- a/server/router/driver.ts
+++ b/server/router/driver.ts
@@ -42,20 +42,24 @@ router.get('/available', validateUser('User'), (req, res) => {
     res.status(400).send({ err: 'startTime must precede endTime' });
   }
 
+  // Condition for conflicting rides, return true if there is a conflict
   const condition = new Condition('status')
     .not()
     .eq(Status.CANCELLED)
     .and()
     .group((condition) =>
       condition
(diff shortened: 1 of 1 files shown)


tuckerstanley (changes requested):
Tried testing this by creating a ride with the drive "Tucker" from 12:11 - 12:15 for TMR, but then I was able to schedule another ride that overlapped with this one with drive "tucker" as well. This would be a bug correct?
Screen Shot 2022-02-27 at 4 17 17 PM

michaeltorku (changes requested):
Could you add some comments inside and across this endpoint? I think it looks good functionality wise - but it took a long time to digest so some comments would help with readability: especially in the condition and the logic toward the end

fengyuli2002:
> -

fengyuli2002:
The parts above which there is no comment belongs to a earlier PR that I made, which was already merged at the beginning of the semester. Those should be correct.

PR #300, opened 2022-03-02
Fix batchGet function

mg2239:
### Summary 

This pull request fixes the batchGet function, used in the `/api/riders/<id>/favorites` endpoint, which causes the server to crash. Without the `else` block, `batchGet` would try to send two responses, as seen in the error below.

image

### Test Plan 

Make a request to `/api/riders/<id>/favorites`.


tuckerstanley (approved):
Tested this. The endpoint returns the data, with no error. Looks good :)

PR #303, opened 2022-03-08
Fix web notifications

mg2239:
### Summary 

This pull request fixes web push notifications, such as allowing them to pop up on desktop, navigating to the website on click, and displaying the notification body in the Notification component.

### Test Plan 

In `server/util/tests`, run `ts-node notification.test.ts`. Make sure you've logged in as a student and admin to get all possible web notifications. You can also adjust `notification.test.ts` to create other notification types.

The web notification payload had to be changed for this to work, so to test locally you need to run the local server and change the proxy of the frontend.

Only tested on Chrome + Mac, so if you have a different configuration it would be good to test there too. Push notifications aren't supported on Safari.


tuckerstanley (approved):
Tested the script, and confirmed that I received a notification in chrome. Works well, good work.

PR #307, opened 2022-03-11
Fix cancelled ride locations in notifications

mg2239:
### Summary 

This pull request populates the ride locations in the notification payload for a cancelled ride. This makes it so a cancelled ride notification doesn't say "undefined" for each location. 

### Test Plan 

Must be tested with this PR: https://github.com/cornell-dti/carriage-rider/pull/61

### Breaking Changes  


mikaylalin (approved):
Tested using Matthew's phone for Rider and Mikayla's Mac for driver. Cancelled rides showed up properly

PR #308, opened 2022-03-17
Error toast

michaeltorku:
### Summary 
Created an error toast according to the Figma guidelines - It will be used to send user notifications if an action or process returns with an error or is ended incomplete. I used an enum to map the types of toasts.

### Test Plan 
Anywhere in the code that says ToastStatus.SUCCESS, change it to ToastStatus.ERROR and test the functionality to see the error Toast. - Easiest to test at user profile by editing user data.

### Notes 

The next PR on this track needs to add the error toast for each case where it should be displayed.

IMG_1827


tuckerstanley (approved):
Changed one of the toasts from SUCCESS to ERROR, and tested to make sure that when used the red toast that you designed showed up, it did. Looks good.

harrisonchin2523 (approved):
Tested the display of the error toast by forcing the SUCCESS instances to be ERROR, and the error toast did end up displaying correctly. Additionally, the changes that were made to ConfirmationModal fixed the error that was occurring when trying to start the server. Good work!

PR #309, opened 2022-03-19
Togglebutton

andrewschoi:
### Summary
Changed button on student page to a toggle slider
Still needs logic implementation

### DIFF
--- a/frontend/src/components/UserDetail/UserDetail.tsx
+++ b/frontend/src/components/UserDetail/UserDetail.tsx
@@ -112,7 +112,11 @@ const UserDetail = ({
           <div className={styles.userEditContainer}>
             {rider && !isRider ? (
               <label className={styles.switch}>
-                <input type="checkbox" onClick={toggleActive} checked= {rider.active} />
+                <input
+                  type="checkbox"
+                  onClick={toggleActive}
+                  checked={rider.active}
+                />
                 <span className={styles.slider}></span>
               </label>
             ) : null}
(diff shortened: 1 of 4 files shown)


frontend/src/components/UserDetail/UserDetail.tsx
```diff
@@ -111,9 +111,10 @@ const UserDetail = ({
           </div>
           <div className={styles.userEditContainer}>
             {rider && !isRider ? (
-              <Button onClick={toggleActive}>
-                {rider.active ? 'Deactivate' : 'Activate'}
-              </Button>
+              <label className={styles.switch}>
+                <input type="checkbox" />
```
michaeltorku, line 115: We want to be able to control the button in-code as well  - which means we want to have a variable which when true, will set the button to be on when the user loads the page.  We don't want the button to default to false whenever the user opens the page (because we want it to match the state of the user which could be on or off at the time the user visits this page)

michaeltorku (changes requested):
We want to be able to control the button in-code as well  - which means we want to have a variable which when true, will set the button to be on when the user loads the page.  We don't want the button to default to false whenever the user opens the page (because we want it to match the state of the user which could be on or off at the time the user visits this page)

RohitValiveti:
Nice work Andrew! The slider looks clean! Based on the Figma design, maybe on top of the active/inactive status being denoted by the color of the slider, you could add a text/label next to the slider saying "active"/"inactive". Regardless, I think it is intuitive enough to differentiate the status of the student based off the color the status and the alert you get when you press the switch. I just wanted to mention that for consistency sake. Nice work!

michaeltorku:
Terminated - Cause : Duplicated

PR #312, opened 2022-03-20
Format edit student modal

harrisonchin2523:
### Summary 

I formatted the edit student modal so that it would match the Figma screen a little more and removed the extraneous SRLabel import.  
Screen Shot 2022-03-31 at 8 51 31 PM

### Notes 

I am a little unsure if I formatted everything the way that we'd want it to be because the textboxes are meant to stay, but they are not present in the Figma. Changes can easily be made in the future if they are not satisfactory though.


fengyuli2002 (commented):
Good job making the modal looks better! A few minor issues that are unclear to me:
1. I didn't find a figma card about Edit Student. I found one about Add Student and it seems you're following that design. Is that the correct design here?
2. In `frontend/src/components/Modal/ridermodal.module.css`, would changing the margin from percentages to `rem` cause inconsistency across different screen sizes?

Appreciate if you can answer these!

harrisonchin2523:
> Good job making the modal looks better! A few minor issues that are unclear to me:
> 
> 1. I didn't find a figma card about Edit Student. I found one about Add Student and it seems you're following that design. Is that the correct design here?
> 2. In `frontend/src/components/Modal/ridermodal.module.css`, would changing the margin from percentages to `rem` cause inconsistency across different screen sizes?
> 
> Appreciate if you can answer these!

1. I think the Edit Student card is found in the same row as Add Student in the Figma, but just further right in that row.
2. I think you're right about this and I will look into this.

harrisonchin2523:
> Also it seems the `SRLabel` you imported is not used. Any special purposes?

This was something I imported and thought I might use, but I forgot to remove it, I can remove it now.

fengyuli2002:
> > Also it seems the `SRLabel` you imported is not used. Any special purposes?
> 
> This was something I imported and thought I might use, but I forgot to remove it, I can remove it now.

Yeah sure!

fengyuli2002:
> > Also it seems the `SRLabel` you imported is not used. Any special purposes?
> 
> This was something I imported and thought I might use, but I forgot to remove it, I can remove it now.

Thanks, now I see that. It seems everything is great except the black button is "Save" instead of "Edit student". Good job on the front end!

michaeltorku:
@harrisonchin2523 I think there's one outstanding requested change - the black button is "Save" instead of "Edit student". Other than that, I think this PR is good to go.

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -219,7 +221,7 @@ const RiderModalInfo = ({
           Cancel
         </Button>
         <Button type="submit" className={styles.submit}>
-          {isEditing ? 'Edit a Student' : 'Add a Student'}
+          {isEditing ? 'Save' : 'Add a Student'}
```
michaeltorku, line 224: 'Save' should be 'Edit Student'

michaeltorku (changes requested):

RohitValiveti:
Nice job on the reformatting, Harrison! The changes seem to better resemble the Figma. Just for consistency with the Figma, the "to" between the join and end dates seems to be a bit smaller than the Figma and uncentered. Again, this is a bit knit picky but wanted to address it here for consistency sake. Nice work!
Screen Shot 2022-04-16 at 6 28 31 PM

andrewschoi:
Nice work Harrison! I think the 'to' should be more centered between the join and end date. Good job!

PR #314, opened 2022-03-21
Get a rider's active state

fengyuli2002:
### Summary 

Implement a new endpoint `api/riders/id/active` that returns a boolean representing if a rider is active. This is done by scanning the database.

### Test Plan 

I tested it on Postman and it worked fine.

### Notes 

There is a [twin task](https://trello.com/c/pXncfRSS) which I think might not be necessary. TPM worth a look. It's about a post endpoint for a rider's active state.

### DIFF
--- a/server/router/rider.ts
+++ b/server/router/rider.ts
@@ -56,6 +56,16 @@ router.get('/', validateUser('Admin'), (req, res) => {
   db.getAll(res, Rider, tableName);
 });
 
+// Get a rider's active state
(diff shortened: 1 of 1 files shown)


mg2239:
not sure if this endpoint is necessary, since `GET /riders/<id>` already returns if the student is active
@michaeltorku thoughts?

michaeltorku:
Good catch - I think we can stick to using the rider fetch endpoint for this use case

PR #316, opened 2022-03-25
Toggle button completed

andrewschoi:
Changed button to toggle slider and keeps track of rider active state with useState()


tuckerstanley (changes requested):
Good work here, just a few changes. Referencing the figma, there should be text to the left of the actual toggle button which changes when activated vs deactivated. I attached a screenshot below of the figma, and what I am talking about. 
Screen Shot 2022-04-14 at 8 39 28 PM


Also, with the toast which pops up when you switch the button from active -> deactive and the other way, I think you mixed up the functionality (or at least with the toast). When it is on (blue) it should be active, and when it is off (not blue) it should be deactive, as the figma says.

michaeltorku (changes requested):
The functionality looks good. Just one thing: When someone changes from active to inactive for the first time the notification displays the wrong message for a split second, before changing to the correct message => This is because the code sets isShowing to true before updating the toast message, try and do some code restructuring to change the message before showing the toast.

harrisonchin2523 (approved):

harrisonchin2523 (dismissed):
It seems that you have fixed the previous issue of the incorrect error message momentarily being displayed, good work! The text to the left of the button is a little bit off-center but otherwise looks good.

caggarwal8 (approved):
Functionality looks good, as does the new centering and display — great work! I think that everything looks in line with the Figma and that no further changes are needed.

selenaliu1:
close

selenaliu1:
old

PR #317, opened 2022-04-01
Fixing Add Student 

tuckerstanley:
### Summary 

In the PR, I fixed the add students functionality. It wasn't working because it expected an array, but it was receiving a string instead. In order to fix this, I needed to change its type in every instance, and go into the backend and remove all of the old accessibility instances that used arrays instead of strings.

### Test Plan 

In order to test this, you will need to change your proxy to the local host instead of the heroku host. You can do this in frontend -> package.json near the bottom. Just paste the local backend link into proxy. Then run the server, and run the frontend, and try adding a student!
Screen Shot 2022-03-31 at 8 41 17 PM

### DIFF
--- a/server/models/rider.ts
+++ b/server/models/rider.ts
@@ -25,7 +25,7 @@ export type RiderType = {
   lastName: string;
   phoneNumber: string;
   email: string;
-  accessibility: Accessibility[];
+  accessibility?: string;
   organization?: Organization;
   description?: string;
   joinDate: string;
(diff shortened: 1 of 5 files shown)


server/models/rider.ts
```diff
@@ -62,10 +62,8 @@ const schema = new dynamoose.Schema({
     validate: (email) => isEmail(email as string),
   },
   accessibility: {
-    type: Array,
-    schema: [String],
+    type: String,
     required: false,
-    default: [],
   },
```
mg2239, line 67: just for some extra security, add an enum field to accessibility (like organization below)

mg2239 (changes requested):

mg2239 (approved):

PR #322, opened 2022-04-14
Update Rider Web schedule on edit or cancel

mg2239:
### Summary 

When editing and cancelling rides, the schedule on Rider Web now updates automatically.

### Test Plan 

Run the frontend, schedule a ride, and edit/cancel it.


harrisonchin2523 (commented):
When I edit a scheduled ridetime to a different one, the ride that was previously in the scheduled rides part of the website moves to the unscheduled rides part. I'm not sure if that's correct based off of the names. The live editing is working as intended though which is good!

mg2239:
> When I edit a scheduled ridetime to a different one, the ride that was previously in the scheduled rides part of the website moves to the unscheduled rides part. I'm not sure if that's correct based off of the names. The live editing is working as intended though which is good!

that's intended, since if we change a ride, the changes might conflict with the previously assigned driver's schedule

michaeltorku:
I got this error while trying to look at the riders page (while logged in as an admin). Might it be because of the previous accessibility PR?
image

michaeltorku (changes requested):
rider web doesn't allow me to request a ride (it doesn't allow me to select a location)

mg2239:
@michaeltorku just merged master, try repulling and testing again

harrisonchin2523 (approved):
Editing/adding a ride does update automatically

PR #332, opened 2022-04-29
collapsible archive for students

harrisonchin2523:
### Summary 

I created a collapsible component for students that filters active and archived students by mostly following the design from Figma, you can reference "collapsible panels without boxes and black text. 

### Notes 

I changed buttons in FormElements so that when hovered over, they will change their background color. I'm not sure if this is necessary, but Carl gave me the go-ahead and the change is easily reversible.

### DIFF
--- a/frontend/src/components/Collapsible/collapsible.module.css
+++ b/frontend/src/components/Collapsible/collapsible.module.css
@@ -20,7 +20,8 @@
 }
 
 .icon {
-  margin-left: 0.75rem;
+  margin-left: auto;
+  display: block;
 }
 
 .contentContainer {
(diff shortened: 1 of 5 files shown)


RohitValiveti:
Good stuff Harrison, the functionality seems to be working as intended! A couple design differences I noticed from the Figma:
- The categories/personal info (Name/NetId, Number, Address, etc) should be visible prior to expanding the active/archived students: see the review comment I also made
- The component holding the "Active"/"Archived" students should have identical styling as each individual student, per the Figma
I think besides that everything looks good. Great Work!

tuckerstanley (commented):
Let's talk about this during work session, because I feel like this PR conflicts with the Search Bar PR I just made. It might be better to have it look more like one of these two options, where inactive/active is a filter option instead of a set collapsable. 
Screen Shot 2022-05-02 at 4 05 11 PM
Screen Shot 2022-05-02 at 4 06 26 PM
Screen Shot 2022-05-02 at 4 06 26 PM

PR #333, opened 2022-05-01
adjusted info container height and made scrollable

RohitValiveti:
### Summary 

Adjusts Employee Card sizing, to be a fixed height rather than larger/smaller depending on timings. To view other times one can scroll

### DIFF
--- a/frontend/src/components/Card/card.module.css
+++ b/frontend/src/components/Card/card.module.css
@@ -41,6 +41,8 @@
 .infoContainer {
   display: flex;
   align-items: center;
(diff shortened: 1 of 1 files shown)


andrewschoi (commented):
Looks good Rohit!

tuckerstanley (changes requested):
Looks pretty good, the height is definitely better. From looking at the trello task, it mentions "set a generic image for each box and replace the poorly displayed image with those", are you going to incorporate this into this PR?

PR #334, opened 2022-05-02
Upcoming rides

andrewschoi:
added a upcoming rides card to student


tuckerstanley (changes requested):
Hey Andrew!  It seems as if you have a lot of repeated code from your other PR, the toggle button PR, which hasn't been approved yet into the code base. Before we review this, can you delete the repeating code from this branch? It should just be in the other branch for now. Once you do that, just push the changes, and I can review this PR.

PR #335, opened 2022-05-02
Created Search Bar and all Search Functionality for Students

tuckerstanley:
### Summary 

As said in title, created search bar and all search functionality for students. Once any changes are made to this PR, I will do a similar process for employees. 

### Test Plan 

You can just test by experimenting with the search feature and making sure it has the proper functionality, as well as that the formatting looks good.


harrisonchin2523 (approved):
The search bar looks good! There may be future work where we could account for if the user adds extraneous white space, but it may not be necessary anyway.

PR #349, opened 2022-09-29
Changed phone input on edit student to type tel

andrewschoi:
-Changed phone input type on edit student -> edit student to type tel


harrisonchin2523 (changes requested):
When I try to add a rider without needs, it doesn't allow me to add them, nor does it give a reason as to why. An error message may be needed in order to inform users of this, unless this is unintentional?

RohitValiveti:
When I have an input in the phone number field, but is not the proper length of a phone number, the error message still shows "Phone number cannot be empty". It might make sense to have a different error message for when the field is empty vs when the input type is not valid.
Screen Shot 2022-10-06 at 4 13 07 PM

pratyush1712 (approved):
Good work! All the requested changes were addressed. I just removed console.log (they should not be merged in master). Otherwise, it looks good to me! @harrisonchin2523 could you look over the updated files and see if you can approve the changes

PR #356, opened 2022-10-13
Patched home redirect on refresh for admin

andrewschoi:
-added local storage for each page url, on reset redirects to last page

### DIFF
--- a/frontend/src/components/AuthManager/AuthManager.tsx
+++ b/frontend/src/components/AuthManager/AuthManager.tsx
@@ -50,7 +50,7 @@ export const AuthManager = () => {
 
   function logout() {
     signOut();
-    // localStorage.removeItem('userType');
+    localStorage.removeItem('userType');
     localStorage.removeItem('lastPage')
     if (jwt) {
       setJWT('');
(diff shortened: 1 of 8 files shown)


pratyush1712 (commented):
Good Work, Andrew! Seems like the refresh, and redirect thing works perfectly on the admin site. However, when I try to log in to the student's account after I have logged out of the admin account, the page keeps refreshing. I think that is because the last time I was on the admin page, lastPage was set to something, and after login out, I am trying to login to the student's account but the AuthManager will try to redirect me to the last page, which I can't access any more because student account doesn't have access to the admin endpoints. So, it redirects to the login page again, and then it's an infinite loop. The only way I can get out of the loop is by manually removing the last page from the local storage. This is just one bug I found, otherwise, this PR looks good!

pratyush1712 (commented):
I think the recent changes that were merged in (new authentication flow) have addressed the issue of redirects. I doubt we need to store the info about the last page in the local storage anymore. However, I liked the idea of utilizing local storage to make sure the user gets redirected back to the last page! Good Job!

PR #358, opened 2022-10-17
Added Page Titles

epicdragon44:
### Summary 

Pages have been renamed according to a more descriptive design scheme. Overview pages (that list lots of smaller rows of data, for instance) have been renamed to "[Page] - Carriage." Specific pages (eg. for Carl Chen's Employee page) have been renamed to "[FName LName] - Carriage."

This was done with a simple useEffect on each relevant component that set document.title depending on either the component it's in, or the relevant object (as long as it's !undefined) (eg. rider, employee).

### Test Plan 

N/A - Frontend Only


tuckerstanley (approved):
Good work, this definitely helps with the flow of the app from section to section. LGTM! 🚀

PR #359, opened 2022-10-17
Cancel Ride Button for Admin

RohitValiveti:
### Summary 

Added delete/cancel ride button option on rider admin. Trello [Card](https://trello.com/c/I6FX93Wz/427-add-a-cancel-ride-button) (Figma design linked on card)
Option for deleting single or all of a recurring ride.
Added toast confirmation upon cancellation

Screen Shot 2022-10-17 at 11 45 17 AM

Screen Shot 2022-10-17 at 11 46 01 AM

Screen Shot 2022-10-17 at 11 46 19 AM

### DIFF
--- a/frontend/src/components/Modal/DeleteOrEditTypeModal.tsx
+++ b/frontend/src/components/Modal/DeleteOrEditTypeModal.tsx
@@ -13,6 +13,7 @@ type DeleteOrEditTypeModalProps = {
   onClose: () => void;
   deleting: boolean;
   onNext?: (single: boolean) => void;
+  isRider?: boolean;
 };
 
 const DeleteOrEditTypeModal = ({
@@ -21,6 +22,7 @@ const DeleteOrEditTypeModal = ({
   onClose,
   deleting,
   onNext,
+  isRider,
 }: DeleteOrEditTypeModalProps) => {
   const [single, setSingle] = useState(true);
   const { withDefaults } = useReq();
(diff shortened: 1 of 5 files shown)


tuckerstanley (changes requested):
This is a great start Rohit! I did some manual tests on it to test functionality, and they were working well. One thing that I noticed is that the ride doesn't get deleted immediately, it requires a page refresh. Also, the modals do not completely match the figma description, this first one contains a header and X that the figma doesnt have.

Screen Shot 2022-10-18 at 2 27 57 AM Screen Shot 2022-10-18 at 2 36 12 AM

For this second one, there are a few things different, as well as no spacing between the buttons and the button being filled in black. You can reference the CSS portion of Figma to help with this. 

Screen Shot 2022-10-18 at 2 33 09 AMScreen Shot 2022-10-18 at 2 38 16 AM

tuckerstanley (changes requested):
The design of the modal looks much better! It matches the figma now, which is great. This PR is looking pretty good, though there is one more thing to fix.  When i mentioned the page requiring a refresh to update the ride not being there, I didn't mean to do like a hard reload as you did. The page shouldn't reload when a ride is deleted, but it shouldnt need to be refreshed manually either, the ride should just be deleted in the moment.

It might be helpful to look at the code for when rides are added to see how that is done in the moment without requiring a refresh!

tuckerstanley (approved):
Great work! LGTM!! 🚀🚀

PR #360, opened 2022-10-18
Update Collapsible to comply with Accordion ARIA

helenxuyang:
### Trello tasks
[Accordion lacks expected WAI-ARIA markup](https://trello.com/c/Fa8piCvR/399-accordion-lacks-expected-wai-aria-markup)
and
[Accordions contain an ambiguous See More button that lacks description of its purpose](https://trello.com/c/KLpNgEwz/400-accordions-contain-an-ambiguous-see-more-button-that-lacks-description-of-its-purpose)

### Summary 

This pull request is the first step towards improving the screenreader accessibility of the Collapsible component to comply with the ARIA specs for Accordions (https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion.html)

- [x] Move button role to the header div with the title and button
- [x] Add aria-expanded and aria-controls according to ARIA specs
- [x] Add tabindex -1 and aria-hidden on icon button to avoid duplicate buttons since the header div acts as the button
- [x] Add role of region to the content of the Collapsible according to ARIA specs   

### Test Plan 

Tested with NVDA:
1. Focus on Your Upcoming Rides: "Your Upcoming Rides, heading 2, button collapsed"
2. Toggle with space or enter: "Expanded" / "Collapsed"

### Notes 

* Need help testing with VoiceOver from someone with a Mac
* The ID being 'region' + the title of the section is a little questionable, but I need a unique ID and I'm guessing the title will always be unique; if this isn't the case then we can add ID as a parameter and always set it ourselves

### Breaking Changes  

### DIFF
--- a/frontend/src/components/Collapsible/Collapsible.tsx
+++ b/frontend/src/components/Collapsible/Collapsible.tsx
@@ -12,7 +12,8 @@ const Collapsible = ({ title, children }: CollapsibleSection) => {
   const icon = expanded ? down : up;
 
   const handleKeywordKeyPress = (e: React.KeyboardEvent) => {
-    if (e.key === 'Enter') {
+    e.preventDefault();
+    if (e.key === 'Enter' || e.key === ' ') {
       setExpanded(!expanded);
     }
   };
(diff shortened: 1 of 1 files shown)


tuckerstanley (changes requested):
Looks pretty good! We just tested it to confirm that it works on mac, but we need to fix the keyboard navigation which is messed up because of this change.  Seems like you have a good idea how to fix it!

helenxuyang:
@tuckerstanley Added Enter and Space for button role (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#examples)

selenaliu1:
old

PR #366, opened 2022-10-27
revamping forms in fieldsets and legends

pratyush1712:
### Summary

This pull request is the first step towards improving the structure of forms using fieldset and legends

- finished revamping the code for "add a student" form. let me know how it looks and will use the same structure for "add a employee" form as well


harrisonchin2523 (commented):
For this modal, there seems to be a lot of extra space, should we trim all of that whitespace out?

tuckerstanley (changes requested):
Screen Shot 2022-11-16 at 12 26 21 PM
Can we pin address to the row below? on my computer, the label and input are on different lines. Otherwise, it looks very nice!

SGupta101 (commented):
Good job with this PR. Add a Student looks good, but I think the amount of white space at the top of the "General Info" and "Duration" boxes are different than the bottom. I also agree that there seems to be a lot of white space that could be trimmed. I think the "Needs" drop-down could also be more centered (height-wise) between the two boxes (currently it is closer to the "General Info" box).

SGupta101 (commented):
Good job with this PR! The changes overall look good, however, I think that some of the spacing could be fixed. There seems to be more white space at the top of the "General Info" and "Duration" boxes than the bottom. The "Needs" dropdown is also closer to the "General Info" box and should instead be centered (height-wise) between the two surrounding boxes.

PR #368, opened 2022-10-28
Augmenting the rider profile page

harrisonchin2523:
The functionality of the rider profile page should work now, past and upcoming rides should be in the correct collapsibles now.


frontend/src/pages/Rider/Schedule.tsx
```diff
@@ -5,54 +5,121 @@ import React, {
...
+          <RiderScheduleTable data={pastRides} isPast={false} email={email} />
+        </Collapsible>
+      )}
+
+      <label>
+        <input
+          type="checkbox"
+          checked={filter}
+          onChange={() => setFilter(!filter)}
+          style={{
```
tuckerstanley, line 111: This should be moved into a seperate CSS file

tuckerstanley (changes requested):
Looks pretty good so far! 

Some comments:
- The figma mock has spacing inbetween the rider information and the collapsables, which your code doesn't have. 
- CSS should go into CSS file
- What is your testing plan for how outcome changes pending a ride being completed / attended? Also what are the other options then "requested" here?
- The headers for different days should be removed. This can probably be done by passing some sort of information into the schedule table component, and having all of the headers render on a condition. 
- The figma has the "show absent rides" checkbox above where the rides get displayed in the collapsable. Personally, I think the way you made it with it below makes sense, and seems much easier to implement, but lets talk to Larrisa about it. 

Great work so far, LMK if you have any questions about my comments.

frontend/src/pages/Rider/Schedule.tsx
pratyush1712: second that! maybe you could create a CSS file if not already exists because random blocks of styling code in the main files is probably not good practices

pratyush1712 (commented):
The PR looks solid and it seems to work properly! Good job! As Tucker suggested could you maybe try adding some test plan for the changes made?

namanhboi:
close pr

PR #370, opened 2022-11-02
Minutes asc269

andrewschoi:
When visiting the Employees page on Admin, users can now see driver's availability with minutes

### DIFF
--- a/frontend/src/components/EmployeeModal/EmployeeModal.tsx
+++ b/frontend/src/components/EmployeeModal/EmployeeModal.tsx
@@ -59,9 +59,6 @@ const EmployeeModal = ({ existingEmployee }: EmployeeModalProps) => {
   const modalTitle = existingEmployee ? 'Edit Profile' : 'Add an Employee';
   const submitButtonText = existingEmployee ? 'Save' : 'Add';
 
-  React.useEffect(() => {
-    // console.log(drivers)
-  }, []);
 
   const openModal = () => {
     setIsOpen(true);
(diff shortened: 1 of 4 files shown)


frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -53,12 +53,13 @@ const EmployeeModal = ({ existingEmployee }: EmployeeModalProps) => {
   );
   const [imageBase64, setImageBase64] = useState('');
   const { withDefaults } = useReq();
-  const { refreshAdmins, refreshDrivers } = useEmployees();
+  const { refreshAdmins, refreshDrivers, drivers } = useEmployees();
```
tuckerstanley, line 56: What do you use drivers for? It doesnt seem like it is used in employeeModal at all.

frontend/src/components/UserDetail/EmployeeDetail.tsx
```diff
@@ -127,6 +127,7 @@ const findEmployee = (
   const employee = DriverToEmployees(drivers).find(
     (employee) => employee.id === employeeId
   );
+  console.log(drivers);
```
tuckerstanley, line 130: Delete! We don't want console.logs getting merged in, they should just be used for testing purposes.

tuckerstanley (changes requested):
Looking good! See my comments, but these should be quick fixes! Nice job.

epicdragon44 (approved):
Looks like all of Tucker's issues were addressed. Approving for merge.

PR #373, opened 2022-11-10
Auth Flow

pratyush1712:
## Summary

This pull request is the first step toward implementing a better authentication flow using up-to-date libraries.

## Trello Task
<blockquote class="trello-card"><a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ksJabM5Q&#x2F;441-updating-google-authentication">Updating Google Authentication</a></blockquote>

While working on the Trello task, I realized that the current authentication flow is not optimal and there could be some issues in the future. Therefore, I made a bunch of changes:
<ol>
<li> Migrated from react-google-login to react-oauth/google. <br />
<b>Reason:</b> The old library is deprecated.<br />
<b>Changes Made:</b> The new library enables us to verify the user and the token generated when the user selects an account from the client side itself. Therefore, we don't have to verify the token on the server anymore.

Hence, the functionality of this function,
https://github.com/cornell-dti/carriage-web/blob/cfb45d13d91e2b8431de1f04626196c8741f40cb/server/router/auth.ts#L22-L25

was implemented using,
https://github.com/cornell-dti/carriage-web/blob/d321f9db4103394446930d89019025bf843d1a53/frontend/src/components/AuthManager/AuthManager.tsx#L89-L93</li>
<li> Storing user data like server jwt token, user type, user info, and user id in local storage or cookies <br />
<b>Reason:</b> Earlier we were making requests for a new jwt token to the server every time the browser session ends or the page is refreshed which could possibly compromise the budget as there are a lot of unnecessary requests directed to AWS dynamodb.<br/>It also causes the client side to act a little weird on each refresh (current page $\rightarrow$ login page $\rightarrow$ home page.) We could definitely have a better flow (current page $\rightarrow$ current page) if we don't have to login every time there is a refresh.<br />
<b>Changes Made:</b>
<ol>
<li> JWT tokens received from the server are now stored on the client side using cookies. However, storing a token that could potentially give anyone access to the entire database is very risky, and therefore the tokens are encrypted before being added to the cookies so that even if someone gains access to the token they can't make any requests to the backend (thanks to @epicdragon44 for suggesting encryption.)</li>
<li> User id, user type, and user info are stored in the local storage. By doing this, we are removing the necessity to fetch these details from the server when the user ends the browser session or refreshes the page.</li>
</ol>
</li>
</ol>

## Video description
[Watch the video](https://imgur.com/HUY6ROZ.mp4)


epicdragon44 (changes requested):
Hey Pratyush! Great PR, and fantastic documentation in a very detailed PR description.

A few small changes I'd implement.

- If avoidable, remove `axios` and switch back to using `fetch` for HTTP clients. Let's not import behavior that can work with vanilla, provided technologies. After switching, test all your fetch requests and make sure the functionality is still the exact same (there's a few differences between axios/fetch that you should be aware of).
- Minimize your use of the `any` type if possible.

One last thing: for the sake of consistency, I committed a few type fixes:
- Declare a ValidUser type that's either Admin or Rider to be used in >1 files.
- Optional parameters with question marks `?` don't need an additional `| undefined` type indicator, and some parameters with `| undefined` can be converted to use `?` instead.
These are up to you for your review and revision as well.

tuckerstanley (approved):
Great work on this PR!! Thanks for walking me through it, it seems like you addressed all of Daniel's feedback and after removing the console.log your code looks very clean. Documentation looks great as well. Since you merged the bug fix PR into here, you wont need to merge that one as well. LGTM! 🚀🚀

PR #376, opened 2022-11-18
fixed api

pratyush1712:
### Summary

Implemented a quick fix for the rides API. Apparently, the populate function is not working properly. However, we can just convert the data fetched from the db to JSON and return that.

https://github.com/cornell-dti/carriage-web/blob/a4b3a3b17121747ba733f2a180c004b89032b973/server/router/common.ts#L188-L190

### Test Plan
Data seems to be fetched properly here
image

### DIFF
--- a/frontend/src/components/AuthManager/AuthManager.tsx
+++ b/frontend/src/components/AuthManager/AuthManager.tsx
@@ -77,7 +77,7 @@ export const AuthManager = () => {
         },
       })
         .then((res) => res.json())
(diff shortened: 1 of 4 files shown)


harrisonchin2523 (commented):
Screenshot 2022-11-20 at 1 29 13 PM

When experimenting with your changes, it seems like they function initially, but when I tryt o add a ride, I'm faced with these errors. From here, I can't add a ride or log back in unless I restart the server.

PR #380, opened 2022-11-22
Added aria-required=true

pratyush1712:
### Trello Task
<blockquote class="trello-card"><a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;FY7Afy1m&#x2F;418-mandatory-form-inputs-are-not-identified-with-either-alt-or-aria">Mandatory form inputs are not identified with either ALT or ARIA</a></blockquote>

#### Can see the aria-required=true
image


frontend/src/components/EmployeeModal/EmployeeInfo.tsx
```diff
@@ -29,6 +29,7 @@ const EmployeeInfo = ({
           type="text"
           defaultValue={firstName}
           className={cn(styles.input)}
+          aria-required={true}
```
epicdragon44, line 32: Looks good to me! Looks like you covered all the inputs we need covered. 

A light suggestion:

epicdragon44, line 32:
```suggestion
          aria-required="true"
```

A light suggestion: use a string "true" instead of a JavaScript boolean. You're passing a pretty standard HTML parameter, so let's make the JSX match that instead of starting to mix types unnecessarily.

epicdragon44 (dismissed):
Great work overall! Looks like you covered all the inputs we need covered. 

I'll go ahead and approve. Make the optional suggested change if you want.

frontend/src/components/EmployeeModal/EmployeeInfo.tsx
pratyush1712: yup sounds good. thanks for the suggestion!

epicdragon44 (approved):
Perfect! The only issue I pointed out was addressed. Ready for merge!

PR #398, opened 2023-02-09
Switch 'activated' and 'deactivated' text in toast

Enochen:
### Summary 

Previously when you select activate, a toast will show "Rider deactivated" and vice versa. It was flipped and this fixes that.

Corrected behavior:
image

image


harrisonchin2523 (approved):
The changes look good to me! Nice job fixing up this small bug and welcome to Carriage.

PR #399, opened 2023-02-09
change redirect based on sign in status with '/'

harrisonchin2523:
### Summary 

Before the changes, regardless of sign-in status, the user would be redirected to the login page if their path was "/". With these changes, the user should be redirected to their respective home pages whether they are an admin or a student.


RohitValiveti:
Nice work Harrison. The routes seem to be working as intended from my manual testing.

RohitValiveti (approved):
Nice work Harrison. The routes seem to be working as intended from my manual testing.

PR #400, opened 2023-02-16
Dockerization

pratyush1712:
### Summary

This pull request is the first step towards dockerization
- [x] Set up `Dockerfile` and `docker-compose.yml` files that containerize an application.
- [x] Implemented a `ci-cd` pipeline to test dockerization on every pull request.
- [ ] Fix any vulnerabilities and follow best coding practices.

### Test Plan
- cd into `root`
- run `docker-compose build` and `docker-compose up`
- head over to `http://localhost:3001/` and Test the platform


frontend/package.json
```diff
@@ -32,8 +34,8 @@
     "typescript": "^4.8.4"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "env-cmd -f ../.env react-scripts start",
```
Enochen, line 37: Is this change meant for local dev or deployment?
If we want to include environment variables in Github Actions, we can add our secrets to the repo settings instead.

package.json
```diff
@@ -22,7 +22,7 @@
   "scripts": {
     "heroku-postbuild": "npm run build:frontend && npm run build:server",
     "start:frontend": "cd frontend && npm run start",
-    "start:server": "cd server && npm run start",
+    "start:server": "cd server && npm run dev",
```
Enochen, line 25: What's the reason for this change?

server/package.json
```diff
@@ -32,7 +32,8 @@
     "web-push": "^3.4.4"
   },
   "scripts": {
-    "start": "nodemon --exec \"ts-node\" app.ts",
+    "start": "ts-node app.ts",
```
Enochen, line 35: In production environments I think `tsc` and `node` is preferred over `ts-node` because of performance reasons. If sticking to `ts-node` it's better to use `ts-node --transpile-only` to skip type processing.

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Install dependencies for the server
+WORKDIR /app/server
+RUN npm install
+
+# Expose port 3001 for the server
+EXPOSE 3001
+
+# Start the server
+CMD ["npm", "start"]
```
Enochen, line 26: Is there already code in the server that serves the built frontend? I don't see it in this diff.

frontend/package.json
pratyush1712: I just made these changes to have just one .env file in root because I was not sure how to make different .env files for frontend and server available through GitHub secrets

package.json
pratyush1712: for development, we would be using nodemon. However, there were issues with using nodemon in deployment

server/package.json
pratyush1712: hmm makes sense... i think we should stick to tsc and node then to make sure the types are checked

frontend/package.json
Enochen: GitHub Actions doesn't read from .env files anyways, so I think this is needed for that case.

package.json
Enochen: Gotcha. `start` is generally reserved for execution in prod rather than dev, so it's probably better to introduce a new `dev:server` command if we wanted that.

server/package.json
Enochen: Yeah feel free to check out my branch where I built with `tsc` and ran `app.js`.

pratyush1712:
moved to `yc728/docker` branch

PR #401, opened 2023-02-17
Dockerization Changes

Enochen:
### Summary 

- [x] implemented X
- [x] fixed Y

### Breaking Changes  


Dockerfile
```diff
@@ -5,9 +5,9 @@ FROM node:16-alpine
 WORKDIR /app
 
 # Copy package.jsons first to install
-COPY package.json /app/package.json
-COPY frontend/package.json /app/frontend/package.json
-COPY server/package.json /app/server/package.json
+COPY package.json package-lock.json ./
```
Enochen, line 8: This is equivalent, but it might fit better with the other COPY commands to move it into `/app/` rather than `./`. Or we can change the others to `./frontend`, for example.

pratyush1712 (approved):
Awesome work! Thanks for improving the architecture and Dockerfiles! I have tested it both locally and by running `docker-compose up --build`

PR #405, opened 2023-02-21
fixed undefined rider name

harrisonchin2523:
### Summary 

There was an issue before where the rider's name would not be displayed on the home page when viewing all of the rides scheduled for the day. Instead of their name, only "undefined undefined" was there; these changes should fix this.
Previous:
Screenshot 2023-02-21 at 6 59 03 PM
With New Changes:
Screenshot 2023-02-21 at 6 58 00 PM

### DIFF
--- a/frontend/src/components/RideModal/RideModal.tsx
+++ b/frontend/src/components/RideModal/RideModal.tsx
@@ -164,8 +164,6 @@ const RideModal = ({ open, close, ride, editSingle }: RideModalProps) => {
         };
       }
 
-      console.log(rideData.rider);
-
       if (ride) {
         // scheduled ride
         if (ride.type === 'active') {
--- a/frontend/src/components/RideModal/RideModal.tsx
+++ b/frontend/src/components/RideModal/RideModal.tsx
@@ -164,7 +164,7 @@ const RideModal = ({ open, close, ride, editSingle }: RideModalProps) => {
         };
       }
 
-      console.log(rideData);
+      console.log(rideData.rider);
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/RideModal/RideModal.tsx
```diff
@@ -164,7 +164,7 @@ const RideModal = ({ open, close, ride, editSingle }: RideModalProps) => {
         };
       }
 
-      console.log(rideData);
+      console.log(rideData.rider);
```
pratyush1712, line 167: pls remove console.logs

pratyush1712 (commented):
Some of the functionalities seems to be broken here: The added ride doesn't show up anywhere
https://user-images.githubusercontent.com/80096811/220504656-43813ce1-c2ac-40df-a85b-0b5f4241d68d.mp4

pratyush1712:
This bug seems to be a lot bigger than I expected and is affecting a lot of parts of the frontend. The root cause of the problem is in server: `/server/src/common.ts`. Last semester, `doc.populate()` was changed to `doc.toJSON()` because Dynamoose made some changes in the `populate` function and made it asynchronous. I thought we are not really using the populate function, so I switched to toJSON. However, this broke some parts of the application, including this one. So, instead of removing populate, the best course of action was to put it back and make the calls asynchronous and use `await` everywhere. I will close this PR for now and make another PR fixing these issues.

It's also a good learning opportunity for everyone lol. We can understand the proper usage of populating data.

PR #406, opened 2023-02-24
Fixed Inactive students

Atikpui007:
### Summary 

- Show Active students on default
- Show Both Active and Inactive students when box is checked

- [ ] implemented 'Show Inactive Students checkbox
- [x] fixed default to only show active students

### DIFF
--- a/frontend/src/components/UserTables/StudentsTable.tsx
+++ b/frontend/src/components/UserTables/StudentsTable.tsx
@@ -34,7 +34,7 @@ const StudentsTable = ({ searchName }: studentTableProps) => {
     '',
   ];
   const [usage, setUsage] = useState<UsageType>({});
-  const [filter, setFilter] = useState(false);
+  const [showInactive, setShowInactive] = useState(false);
 
   useEffect(() => {
     fetch('/api/riders/', withDefaults())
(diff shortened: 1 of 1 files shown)


frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -86,18 +86,18 @@ const StudentsTable = ({ searchName }: studentTableProps) => {
         <Row header colSizes={colSizes} data={headers} />
 
         {(filter
-          ? riders.filter(
+          ? riders.filter((r) =>
+              (r.firstName + ' ' + r.lastName)
+                .toLowerCase()
+                .includes((searchName + '').toLowerCase())
```
Enochen, line 92: There's repeated logic here with the filtering, let's refactor that

Enochen (changes requested):

frontend/src/components/UserTables/StudentsTable.tsx
Atikpui007: Got it. Sending a new request to patch this

Enochen: No need to make a new PR, you can make the changes to the current branch.

frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -66,6 +66,15 @@ const StudentsTable = ({ searchName }: studentTableProps) => {
   const formatDate = (date: string): string =>
     moment(date).format('MM/DD/YYYY');
 
+  const filterStudents = (filter: boolean) =>
```
Enochen, line 69: The name `filter` doesn't really describe the function it's providing anymore, now that we're reducing the filtering when the checkbox is on. Let's give it a more appropriate name.

Enochen, line 69: Also, the parameter isn't necessary as we are not going to call this multiple times, and we have everything we need in scope to calculate the result immediately.

ie `const filteredStudents = riders.filter...`

Enochen (changes requested):

frontend/src/components/UserTables/StudentsTable.tsx
Enochen: This comment is talking moreso about the state name from the `useState` call earlier in the file, not this parameter in particular.

Atikpui007: Got it. I am thinking the the function works the `filter useState `name should be changed to something along the lines of `showInactive` as that is the main purpose of the boolean

Enochen: Yep, that's a great name

Atikpui007: Alright. Is there a specific convention that the codebase follows? Arrow functions vs regular functions? Explicit type naming, etc?
eg 
`const filteredStudents = ....` vs `const filteredStudents = () => ....`

Enochen (approved):
Awesome, thanks for the additional refactoring

PR #414, opened 2023-03-03
correctly filter past/upcoming rides for rider

harrisonchin2523:
### Summary 
There was an issue with the rider aspect of web where past rides would not filter correctly and would still be displayed in the "Your Upcoming Rides" tab. These changes should fix this.

### Notes 
There seems to be an issue with the isPast condition when filtering rides between Schedule.tsx and RiderScheduleTable.tsx. It is not correctly filtering rides that have come to pass. I'm not exactly sure why this is happening, but I wanted to implement this fix so we could correctly filter rides and come back to this after implementing testing.

### DIFF
--- a/frontend/src/pages/Rider/Schedule.tsx
+++ b/frontend/src/pages/Rider/Schedule.tsx
@@ -17,7 +17,10 @@ import styles from './page.module.css';
 
 const Schedule = () => {
   const componentMounted = useRef(true);
-  const [rides, setRides] = useState<Ride[]>();
+  const now = new Date().toISOString();
+  const [rides, setRides] = useState<Ride[]>([]);
+  const [currRides, setCurrRides] = useState<Ride[]>([]);
+  const [pastRides, setPastRides] = useState<Ride[]>([]);
   const { id, user } = useContext(AuthContext);
   const { withDefaults } = useReq();
   document.title = 'Schedule - Carriage';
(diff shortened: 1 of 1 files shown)


pratyush1712 (commented):
Thank you for noticing this issue and fixing it. Can you also share some screenshots or screen recordings explaining the problem and how this pr will fix it? Thanks

harrisonchin2523:
@pratyush1712 here are two screen recordings of the before and after:
- Before:
https://user-images.githubusercontent.com/88568615/222927753-695562f1-3339-48b4-b927-8e54eadac22c.mov
- After:
https://user-images.githubusercontent.com/88568615/222927768-f869ee0d-64ec-4a88-9bc7-ee0252eb0606.mov

frontend/src/pages/Rider/Schedule.tsx
```diff
@@ -17,7 +17,10 @@ import styles from './page.module.css';
 
 const Schedule = () => {
   const componentMounted = useRef(true);
-  const [rides, setRides] = useState<Ride[]>();
+  const now = new Date().toISOString();
+  const [rides, setRides] = useState<Ride[]>([]);
```
pratyush1712, line 21: Good work! These changes work great!

pratyush1712 (approved):
Good job! Thanks for noticing the issue and fixing it!

PR #416, opened 2023-03-06
Fix tsconfig include

Enochen:
### Summary 

When adding a test framework to the server, we mistakenly included the tests folder into the `include` section of the tsconfig.

This causes problems because it nests our `src` folder into the build folder, rather than flattening it out. This causes relative paths to break, resulting in the frontend not being served.

We also do not need to compile the tests because they are not run in prod.
### Test Plan 

Ensure `npm run test` within `server` folder still works.

Ensure `npm build` within `server` folder builds `build` folder without `src` subfolder.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -32,7 +32,7 @@
     "web-push": "^3.4.4"
   },
   "scripts": {
-    "build": "tsc",
+    "build": "tsc --project tsconfig.build.json",
     "start": "node build/app.js",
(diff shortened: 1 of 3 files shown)


pratyush1712 (commented):
I believe there should be a separate tsconfig in the test folder that extends the tsconfig in server to prevent any warnings in the test files

Enochen:
That makes sense, I've added a new `tsconfig.build.json` which excludes the tests folder for building.

PR #417, opened 2023-03-06
Add CD workflow to push container to Docker Hub

Enochen:
### Summary 

When a PR gets merged into master (or main, if we ever switch), we want to build our Docker container and push to the Docker Hub registry.

This diff implements a CD workflow that does that.

### Test Plan 

The workflow has been run successfully.

The new Docker image shows up on the Docker Hub website.

I can pull and run the container locally:
image

The web app is served correctly by the container, and I am able to log in and perform various actions.

### Notes 

Whenever we end up switching to the `main` branch, the reference to `master` will no longer be required and should be removed.

### DIFF
--- a/.github/workflows/cd-workflow.yml
+++ b/.github/workflows/cd-workflow.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - enoch/dockerhub
+      - main
 
 jobs:
   push_to_registry:
--- a/.github/workflows/cd-workflow.yml
+++ b/.github/workflows/cd-workflow.yml
@@ -1,0 +1,25 @@
+name: CD
+
+on:
+  push:
+    branches:
+      - master
+      - enoch/dockerhub
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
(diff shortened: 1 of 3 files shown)


.github/workflows/cd-workflow.yml
```diff
@@ -0,0 +1,25 @@
+name: CD
+
+on:
+  push:
+    branches:
+      - master
```
pratyush1712, line 6: Why was main added here? I think main is way behind the current progress, and we can just merge it in master to get rid of it

Enochen: Sorry, I had the impression that we were eventually going to switch to `main` as the main branch. I can remove it if that is not the case.

pratyush1712 (commented):
I ran into some issues with running the docker image. Could you please look into it again and push the correct image? CULift people might be having issues rn if they pull the image now
image

pratyush1712 (approved):
CULift won't be able to perform Accessibility Tests until late June. So, I think this is good to merge

pratyush1712:
merged!

PR #418, opened 2023-03-07
Runtime secrets refactoring

Enochen:
### Summary 

Previously, our `docker-compose.yml` loaded up various environment variables and made them available to both the build process and the container. Some of these env vars are secrets and are not necessary to the build process, and so we would not like to embed them into our image, which poses as a potential security vulnerability.

We made a change to the Dockerfile in #417 which now expects REACT_APP env vars in the form of build_args. This new change updates the `docker-compose.yml` to supply those args.

This change now also moves these secrets to be loaded from an .env file (so we don't need to explicitly specify each one) into the container using the `env_file` option.

### Test Plan 

`docker-compose build` and `docker-compose up` works as expected.

### Notes 

Right now React-related env vars are still loaded up during build time, but we may actually want to look for a solution that can load during runtime instead.

### Breaking Changes  

### DIFF
--- a/.github/workflows/cd-workflow.yml
+++ b/.github/workflows/cd-workflow.yml
@@ -10,8 +10,8 @@ jobs:
     name: Push Docker image to Docker Hub
     runs-on: ubuntu-latest
     steps:
-      - name: Check out the repo
-        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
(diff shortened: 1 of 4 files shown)


pratyush1712 (approved):
lgtm! make sure to merge the cd pr first, just so that there are no merge issues

pratyush1712:
merged!

PR #419, opened 2023-03-07
Fix Popup tags in production

pratyush1712:
### Summary
My recent changes to remove aria-describedby introduced some bugs in the production. This pr fixes those issues by using component ID instead of hard-coding component's class name

### Test Plan
The `aria-descibedby` tag is gone in production

image


frontend/src/components/Notification/Notification.tsx
```diff
@@ -62,15 +62,15 @@ const Notification = () => {
...
+  useEffect(() => {
+    const element = document.getElementById('popup');
+    element?.removeAttribute('aria-describedby');
+  }, []);
 
   return (
     <Popup
       trigger={
-        <button className={styles.bell}>
+        <button id="popup" className={styles.bell}>
```
Enochen, line 73: id's should be unique, but there's no guarantee for this since it's sitting in a component. Maybe use the `useId` hook to give this element a unique id?

Enochen (approved):
Looks great!

PR #420, opened 2023-03-08
Pratyush/notification design

pratyush1712:
### Summary

Branching off of the previous popup aria tag fix pr, this pr improves the notifications design and implements view more functionality.

### Note
This work is purely intuitive and is not based on any designs, so feel free to suggest any improvements or put it on hold until we have designs ready.

### Test Plan

https://user-images.githubusercontent.com/80096811/223640283-66e9cc35-87d3-4bec-a227-96ac46043b90.mp4


frontend/src/components/Notification/Notification.tsx
```diff
@@ -48,29 +41,18 @@ const Notification = () => {
...
+  useEffect(() => {
+    const element = document.getElementById('popup');
+    element?.removeAttribute('aria-describedby');
+  }, []);
 
   return (
     <Popup
       trigger={
-        <button className={styles.bell}>
+        <button id="popup" className={styles.bell}>
```
Enochen, line 55: I've already commented on this, but we should not be hardcoding id's within components. Maybe we can fix this after we migrate to modern React (18) where you get the `useId` hook natively.

Enochen (dismissed):
Let's fix master right now and reimplement it the right way later

SGupta101 (commented):
Good job with this PR, the design looks great! I was looking at some design inspiration for notifications, and perhaps we could think about adding a "mark all as read" and "delete" notification feature as shown in the example below.

Screen Shot 2023-03-26 at 6 36 11 PM

PR #421, opened 2023-03-08
Update backend to work with mobile

pratyush1712:
### Summary

Made some changes to make the backend compatible with mobile apps. We can't send nested object forms of data through flutter; therefore, we need to stringify the userInfo on the mobile frontend and unwrap it on the backend. We could also keep the backend the same and alter the web frontend - open to suggestions.


RohitValiveti (approved):
Hey Pratyush, the backend still seems to be working fine for me with these changes. I think it may be a good idea to handle the unwrapping on the backend rather than the front end. Personally, I think data handling and refactoring on the backend is easier than on the front end, so it may be less error prone. Otherwise, the PR looks good!

PR #422, opened 2023-03-10
fixed voiceover on modal

RohitValiveti:
### Summary 

When modal is announced, the screenreader properly reads aloud the modal title, instead of 'close button'

### Test Plan 

Opened all the modals and ensured the title was read properly


pratyush1712 (commented):
This announces the "Request a ride button". This is a modal, not a button. It's announcing button because the initial focus when the modal opens is at the close Button, instead, it should be on the Request a Ride Modal component.

Look into the react focus trap documentation and see if there is anything that could be used to change the focus to the entire modal instead of button

PR #423, opened 2023-03-10
Add more Location tests

Enochen:
### Summary 

This diff broadens our testing scope to multiple user types (not just Admins).
It also does a deep comparison of the data returned from the API and the expected data.

### Test Plan 

`npm run test` succeeds locally.

### Notes 

I've only added tests for the base endpoint, more tests are needed for other endpoints like `/:id`

I renamed `locations.test.ts` to `location.test.ts` to match the module we are testing, which is named `location.ts`.

I also added better type annotations to some test-related utils.


pratyush1712 (commented):
looks pretty good! would approve as soon as `:id` params tests are done?

server/tests/location.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  firstName: 'Test-Driver',
+  lastName: 'Test-Driver',
+  availability: {
+    Mon: undefined,
+    Tue: undefined,
+    Wed: undefined,
+    Thu: undefined,
+    Fri: undefined,
+  },
+  vehicle: {
```
pratyush1712, line 29: it should be a `vehicle id` instead of the vehicle object. we are using `populate()` to populate this field. DriverType has `vehicle:VehicleType` but it should be string (the `id`).

Enochen: Do we have types for what request bodies?

Enochen: My mistake, this isn't about request bodies. I mean for what the db structure looks like.

pratyush1712: All the types are in the `src/models`. There are currently no object in the database that actually has the vehicle populated, so we were not getting any errors. However, the current implementation in the types is incorrect. Vehicle field is marked as `VehicleType` but it should be simply `string` because we put the `id` not the entire object.

I asked harrison to fix that in his Driver endpoint test pr.

Unfortunately, we can't use something lie an ObjectId in dynamoose to specify that the `vehicle` is an ObjectId (mongoose has that, but not dynamoose). So, we need to stick with string. I will soon work on fixing all these types issues + updating notion for API

pratyush1712 (approved):
lgtm!

PR #424, opened 2023-03-10
testing rider endpoints

harrisonchin2523:
### Summary 

These are the tests for all the rider endpoints. There seem to be issues with the POST /api/riders/id/favorites and the GET /api/riders/id/favorites endpoints.

### Notes 

I'm not too certain if this is the correct way/style to test backend hits, so let me know!

### DIFF
--- a/server/tests/rider.test.ts
+++ b/server/tests/rider.test.ts
@@ -262,34 +262,6 @@ describe('Testing Functionality of Riders Endpoints', () => {
     });
   });
 
-  const generateGetRiderFavoritesTest = async (authToken: string) => {
-    const res = await request(app)
-      .get('/api/riders/abc-10/favorites')
-      .auth(authToken, { type: 'bearer' })
-      .expect(200)
-      .expect('Content-Type', 'application/json; charset=utf-8');
-    expect(res.status).to.be.equal(200);
-    expect(res.body.data).to.deep.equal(testRiders[0].favoriteLocations);
-  };
-  // testing retrieval of favorite locations
-  // *******************************************************************************
-  // note at the moment I don't think favorites works, it just returns an empty list
-  // *******************************************************************************
-  describe("GET a rider's favorite locations by the rider's ID", () => {
-    it('should return correct response for Admin account', async () =>
-      await generateGetRiderFavoritesTest(adminToken));
-    it('should return correct response for Driver account', async () =>
-      await generateGetRiderFavoritesTest(driverToken));
-    it('should return correct response for Rider account', async () =>
-      await generateGetRiderFavoritesTest(riderToken));
-    it('should fail with 400 given no authorization header', async () => {
-      const res = await request(app)
-        .get('/api/riders/abc-10/favorites')
-        .expect(400);
-      expect(res.body).have.property('err');
-    });
-  });
-
   const newRiderData = {
     email: 'test-email2@test.com',
     phoneNumber: '1234567892',
@@ -351,53 +323,6 @@ describe('Testing Functionality of Riders Endpoints', () => {
       await generateUnauthorizedCreateNewRiderTest(riderToken));
     it('should fail with 400 given no authorization header', async () => {
       const res = await request(app).post('/api/riders/').expect(400);
-      expect(res.body).have.property('err');
-    });
-  });
-
-  // testing the addition of a favorite location for a rider
-  // ****************************************************************************
-  // this test is broken as we cannot retrieve a rider's favorite location list
-  // ****************************************************************************
-  const generateAddFavoriteTest = async (authToken: string) => {
-    const res = await request(app)
-      .post('/api/riders/abc-10/favorites')
-      .send({ id: '2' })
-      .auth(authToken, { type: 'bearer' })
-      .expect(200)
-      .expect('Content-Type', 'application/json; charset=utf-8');
-    expect(res.body.data).to.deep.equal(testLocations[1]);
-    // retrieve this rider's new favorite location list
-    const favoriteLocations = testRiders[0].favoriteLocations;
-    const res2 = await request(app)
-      .get(`/api/riders/abc-10/favorites`)
-      .auth(authToken, { type: 'bearer' })
-      .expect(200)
-      .expect('content-type', 'application/json; charset=utf-8');
-    expect(res2.body.data).to.deep.equal(favoriteLocations);
-  };
-
-  const generateErrorAddFavoriteTest = async (authToken: string) => {
-    const res = await request(app)
-      .post('/api/riders/abc-10/favorites')
-      .send({ id: '2' })
-      .auth(authToken, { type: 'bearer' })
-      .expect(400)
-      .expect('Content-Type', 'application/json; charset=utf-8');
-  };
-
-  describe('POST a new favorite location for a rider', () => {
-    it('should return correct response for Admin account', async () =>
-      await generateAddFavoriteTest(adminToken));
-    it('should fail with 400 given Driver account', async () =>
-      await generateErrorAddFavoriteTest(driverToken));
-    it('should return correct response for Rider account', async () =>
-      await generateAddFavoriteTest(riderToken));
-    it('should fail with 400 given no authorization header', async () => {
-      const res = await request(app)
-        .post('/api/riders/abc-10/favorites')
-        .send({ id: '2' })
-        .expect(400);
       expect(res.body).have.property('err');
     });
   });
--- a/server/tests/rider.test.ts
+++ b/server/tests/rider.test.ts
@@ -200,7 +200,7 @@ describe('Testing Functionality of Riders Endpoints', () => {
         .get('/api/riders/abc-10/usage')
         .auth(authToken, { type: 'bearer' })
         .expect('Content-Type', 'application/json; charset=utf-8');
-      res.status === 200
+      res.status === 200 && expect(res.body).not.have.property('err')
         ? expect(res.body).to.deep.equal(rider0usage)
         : expect(res.status).to.be.equal(400);
     };
@@ -230,7 +230,7 @@ describe('Testing Functionality of Riders Endpoints', () => {
         .get(`/api/rides/${testRides[1].id}`)
         .auth(authToken, { type: 'bearer' })
         .expect('Content-Type', 'application/json; charset=utf-8');
-      res.status === 200
+      res.status === 200 && expect(res.body).not.have.property('err')
         ? (expect(res.status).to.be.equal(200),
           expect(res.body).to.deep.equal(ride.body.data))
         : expect(res.status).to.be.equal(400);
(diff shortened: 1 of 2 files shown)


server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
+import request from 'supertest';
```
pratyush1712, line 1: would recommend renaming file to rider.test.ts to match the source file rider.ts

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    endDate: '2024-03-09',
+    address: '37 Colonial Ln, Ithaca, NY 14850',
+    favoriteLocations: ['west'],
+    organization: 'CULift',
+    photoLink: '',
+    active: true,
+  },
+];
+
+describe('Riders', () => {
```
pratyush1712, line 46: a more descriptive name: Riders Test

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  before(async () => {
+    adminToken = await authorize('Admin', {
+      firstName: 'Test-Admin',
+      lastName: 'Test-Admin',
+      phoneNumber: '1111111111',
+      email: 'test-admin@cornell.edu',
+    });
```
pratyush1712, line 53: why are you not simply using `riders[0]`

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    // add second rider
+    await populateDB(Rider, {
+      id: 'abc-11',
+      email: 'test-email1@test.com',
+      phoneNumber: '1234567891',
+      firstName: 'Test',
+      lastName: 'Testing1',
+      pronouns: 'he/him/his',
+      accessibility: 'Crutches',
+      description: '',
```
pratyush1712, line 80: could use a forEach over `riders` array

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      organization: 'CULift',
+      photoLink: '',
+      active: true,
+    });
+  });
+
+  after(clearDB);
+
+  // testing retrieval of specific rider
+  describe('GET /api/riders/abc-10', () => {
```
pratyush1712, line 94: a more descriptive name: Testing Get Rider by Id

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    it('should return 200 OK', async () => {
+      const res = await request(app)
```
pratyush1712, line 96: more descriptive name: fetch rider with id `abc-10`

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  // testing retrieval of rider profile
+  describe('GET /api/riders/abc-10/profile', () => {
+    it('should return 200 OK', async () => {
+      const res = await request(app)
+        .get('/api/riders/abc-10/profile')
+        .set('Authorization', `Bearer ${adminToken}`)
+        .expect(200)
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      expect(res.status).to.be.equal(200);
```
pratyush1712, line 129: would recommend using `deep`
1. remove the fields from the rider that are not returned and do deep equals expRider

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      expect(res.body).to.have.property('data');
+      expect(res.body.data.reverse()).to.deep.equal(riders);
+    });
+  });
+
+  // testing retrieval of rider profile
+  describe('GET /api/riders/abc-10/profile', () => {
```
pratyush1712, line 122: more descriptive: get rider's profile by id

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        .expect(200)
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      expect(res.status).to.be.equal(200);
+      expect(res.body.organization).to.be.equal(riders[0].organization);
+      expect(res.body.description).to.be.equal(riders[0].description);
+    });
+  });
+
+  // testing retrieval of accessibility info
+  describe('GET /api/riders/abc-10/accessibility', () => {
```
pratyush1712, line 155: again more descriptive

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    it('should return 200 OK', async () => {
```
pratyush1712, line 156: it should fetch accessibility of rider with id abc

server/tests/riders.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        .set('Authorization', `Bearer ${adminToken}`)
+        .expect(200)
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      expect(res.body.data.firstName).to.be.equal('NewName');
+    });
+  });
+
+  // testing the deletion of a rider
+  describe('DELETE /api/riders/abc-10', () => {
+    it('should return 200 OK', async () => {
```
pratyush1712, line 252: check that the object is actually deleted by trying to get the object by id from the model. it should give an error

pratyush1712 (changes requested):
overall, the pr looks good. however, i would recommend using more descriptive test suites names.
also, I don't think that you are testing if the object is actually updated/deleted after a post/put/delete request. you can test them by directly calling the dynamoose functions on the Rider model instead of making a get request

server/tests/rider.test.ts
```diff
@@ -0,0 +1,272 @@
...
+        .set('Authorization', `Bearer ${adminToken}`)
+        .expect(200)
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      expect(res.status).to.be.equal(200);
+      expect(res.body).to.have.property('data');
+      expect(res.body.data.reverse()).to.deep.equal(riders);
+    });
+  });
+
+  const riderProfile = {
```
pratyush1712, line 90: put this inside the it block because it's only used in one test

pratyush1712 (commented):
I see that you are not testing endpoint accessibility by other user types, including Rider and Driver. Some endpoints (where validateUser('User') instead of validateUser("Admin")) can be accessed by all types of users and some are restricted to Admin only. So, make sure the endpoints that Riders and Drivers are authorized to access are accessible and vice-versa. You can get a riderToken and driverToken by basically calling authorize("Rider", testData) in the before

You can look into enoch's location test pr to see how to test for other user types

harrisonchin2523:
> I see that you are not testing endpoint accessibility by other user types, including Rider and Driver. Some endpoints (where validateUser('User') instead of validateUser("Admin")) can be accessed by all types of users and some are restricted to Admin only. So, make sure the endpoints that Riders and Drivers are authorized to access are accessible and vice-versa. You can get a riderToken and driverToken by basically calling authorize("Rider", testData) in the before
> 
> You can look into enoch's location test pr to see how to test for other user types.
> 
> The only reason I am asking for in-depth testing is that Carriage is a pretty large app and is a lot more complicated than what one might expect. Therefore, even a small change in the server can break a lot of stuff under the hood, and we don't want that

how would I know which endpoints can be accessed by other users such as riders and drivers?

pratyush1712:
> > I see that you are not testing endpoint accessibility by other user types, including Rider and Driver. Some endpoints (where validateUser('User') instead of validateUser("Admin")) can be accessed by all types of users and some are restricted to Admin only. So, make sure the endpoints that Riders and Drivers are authorized to access are accessible and vice-versa. You can get a riderToken and driverToken by basically calling authorize("Rider", testData) in the before
> > You can look into enoch's location test pr to see how to test for other user types.
> > The only reason I am asking for in-depth testing is that Carriage is a pretty large app and is a lot more complicated than what one might expect. Therefore, even a small change in the server can break a lot of stuff under the hood, and we don't want that
> 
> how would I know which endpoints can be accessed by other users such as riders and drivers?

If you look into the `server/src/router/rider` you will see some end points have a `validateUser` middleware: it's basically making a check that the user has sufficient `"permissions"` to make the call; this middleware runs before any code in the endpoint is executed. The hierarchy of permission level is like, "User" (kiddo) -> "Driver"/"Rider" (mid) -> "Admin" (the boss). So, if you see a `validateUser("User")` that basically means anyone should be able to access that endpoint; if `validateUser("Admin")` then only admin, etc... Let me know if that makes sense
image

pratyush1712 (changes requested):
please write tests for favorites, currentRide, and usage (similar to stats endpoint in Driver) endpoints. looks good, otherwise!

server/tests/rider.test.ts
```diff
@@ -0,0 +1,587 @@
...
+  });
+
+  // testing retrieval of all riders
+  describe('Testing fetching all riders', () => {
+    const generateGetAllRidersTest = async (authToken: string) => {
+      const res = await request(app)
+        .get('/api/riders/')
+        .auth(authToken, { type: 'bearer' })
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      res.status === 200
```
pratyush1712, line 283: is there a specific reason you are doing a conditional test based on the status? I think it might cause problems in the future. If the endpoint is broken and returns `400 error`, it will still pass the test which we don't want to happen

server/tests/rider.test.ts
```diff
@@ -0,0 +1,587 @@
...
+
+  // testing the retrieval of a rider's favorite locations
+  describe("GET a rider's favorite locations", () => {
+    const generateGetFavoriteLocTest = async (authToken: string) => {
+      const res = await request(app)
+        .get('/api/riders/abc-11/favorites')
+        .auth(authToken, { type: 'bearer' })
+        .expect('Content-Type', 'application/json; charset=utf-8');
+
+      res.status === 200
```
pratyush1712, line 342: same here

server/tests/rider.test.ts
```diff
@@ -0,0 +1,587 @@
...
+        photoLink: '',
+        active: true,
+      };
+      const res = await request(app)
+        .post('/api/riders/')
+        .send(newRiderData)
+        .auth(authToken, { type: 'bearer' })
+        .expect('Content-Type', 'application/json; charset=utf-8');
+
+      res.status === 200
```
pratyush1712, line 490: same here

server/tests/rider.test.ts
```diff
@@ -0,0 +1,587 @@
...
+
+  // testing the updating of info of a rider
+  describe('PUT new information for a rider by id abc-10', () => {
+    const generateUpdateRiderTest = async (authToken: string) => {
+      const res = await request(app)
+        .put('/api/riders/abc-10')
+        .send({ firstName: 'NewName' })
+        .auth(authToken, { type: 'bearer' })
+        .expect('Content-Type', 'application/json; charset=utf-8');
+      res.status === 200
```
pratyush1712, line 526: same here

server/tests/rider.test.ts
```diff
@@ -0,0 +1,587 @@
...
+  });
+
+  // testing the deletion of a rider
+  const generateDeleteRiderTest = async (authToken: string) => {
+    const res = await request(app)
+      .delete('/api/riders/abc-10')
+      .auth(authToken, { type: 'bearer' })
+      .expect('Content-Type', 'application/json; charset=utf-8');
+    res.status === 200
+      ? async () => {
```
pratyush1712, line 561: same here

PR #431, opened 2023-03-21
Stats endpoints Testing

RohitValiveti:
### Summary 

Added Tests for the Stats endpoints (GET and PUT requests)

This pull request regarding the backend testing. [Trello Ticket](https://trello.com/c/M6ZlZQgD/467-stats-endpoint)

### Test Plan 
Test Cases Pass

### Notes 
To Test the GET /api/stats, I generated random from/to dates as query parameters. The response should be stats between the specified dates. I utilized property based testing to ensure the length of the stats elements in the response is equivalent to the number of dates between the randomly generated from_date and to_date. 

To Test the PUT /api/stats, I generated random values for the StatsType and updated the corresponding date. I then tested to make sure that the response of the PUT request (which is the Stats object that was updated), matches the randomly generated data. The helper function to generate the random request data and valid data only only generates 1 at a time and may be generalized to multiple dates in the future, as the PUT request does support updating multiple dates at once.

Finally, I tested the PUT and GET requests but generated a random date to update the stats, and calling the GET request on that same date, and ensuring that the response of the GET request matched the randomly generated stats that was sent with the PUT request.

### Breaking Changes  
Minor changes to Notion API Documentation as request/response formats did not align with code.

### DIFF
--- a/server/tests/stats.test.ts
+++ b/server/tests/stats.test.ts
@@ -108,7 +108,7 @@ describe('Stats Tests', () => {
   });
 
   after(clearDB);
-  
+
   describe('GET /api/stats', () => {
     it('Fetch the stats in the range of specified dates', async () => {
       const res = await request(app)
--- a/server/tests/stats.test.ts
+++ b/server/tests/stats.test.ts
@@ -145,7 +145,7 @@ describe('Stats Tests', () => {
   describe('PUT & GET of same date', () => {
     it('Updates stats at a random Date and GETs Data at that date', async () => {
       const statsData = generateEditDatesData();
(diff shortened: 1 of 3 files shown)


server/tests/stats.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    stats.push(stat);
+  }
+
+  return stats;
+};
+
+// Formats Stats Type Data to proper request structure
+// Currently Assumes stats contains 1 element
+const format_edit_dates_request = (stats: StatsType[]) => {
```
pratyush1712, line 76: please use camelCase only for consistency

Atikpui007:
I run the test on my computer and they all passed. Pretty good

PR #435, opened 2023-03-30
Rides Endpoint Testing

RohitValiveti:
### Summary 

Added Tests for the Rides endpoints (GET, POST, PUT, DELETE requests)

This pull request regarding the backend testing. [Trello Ticket](https://trello.com/c/HwMR2gbo/457-ride-endpoint)

### Test Plan 
Test Cases Pass

### Notes 
Tested all endpoints specified in the notion documentation (i.e. those implemented in the /router/ride.ts file). Also created a data file for all dummy data used in test cases. May be useful to create folder to hold all dummy data across all files in the future.

### Breaking Changes  
Minor changes to Notion API Documentation as request/response formats did not align with code.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -37,7 +37,7 @@
     "dev": "nodemon --exec \"ts-node\" src/app.ts",
     "lint": "eslint . --ext .js --ext .jsx --ext .ts --ext .tsx --fix",
     "type-check": "tsc --pretty --noEmit",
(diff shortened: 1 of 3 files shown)


harrisonchin2523 (changes requested):
The tests look great! The only concern I have at the moment is the absence of actual scheduled rides with a rider and driver. Is there a way you could add these rides and test for their presence? The idea of adding files to be shared across all tests is also a great idea.

PR #438, opened 2023-04-11
consistent employee card sizes

harrisonchin2523:
### Summary 

This task is focused on making all of the employee cards the same size. I set the height of all the cards to 22rem, I am also working on condensing the availability displayed on the cards to reduce the amount of space they take. This is my current progress.

Notes: 
I am trying to space the information displayed on the cards evenly along the vertical axis, but the cards are of display: inline-block which makes this difficult, let me know if anyone knows a way to get around this.

### DIFF
--- a/frontend/src/components/Card/card.module.css
+++ b/frontend/src/components/Card/card.module.css
@@ -1,6 +1,7 @@
 .card {
   display: inline-block;
   width: 17.2rem;
+  height: 23rem;
   background-color: white;
   border-radius: 0.5rem;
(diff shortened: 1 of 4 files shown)


RohitValiveti (approved):
Hey Harrison, the sizing of the cards looks good from my end! They all look the same size. As for the spacing of the content in each card, maybe you could try setting a margin/padding between each piece of information? The exact CSS might be a bit different but thats the first thing that comes to mind

PR #440, opened 2023-04-21
Auth Refactor - Axios and Auth-Code

Enochen:
### Summary 

At some point in the code base, the server was changed to just trust what the client identifies its email as, which is insecure.

This diff uses the `auth-code` OAuth flow to authenticate the client.

This is a smaller-effort change than the original Firebase idea, but it ultimately addresses the same problem without bundling firebase into both the frontend and backend.

This also includes a shift to `axios` from `fetch`. `axios` allows us to set a baseURL "globally" and performs JSON stringification automatically for request bodies. 

### Test Plan 

Login works both with Docker container and running frontend/server separately locally.

### Notes 

The auth client will still prefer the `idToken` form of identification if it is passed in the body. It will ignore the `code` value if `idToken` exists. This feature is important for mobile clients which have a better time getting `idToken`s than web.

### Breaking Changes  

A new environment variable named `REACT_APP_SERVER_URL` was introduced to point the frontend to the right backend url. Note that we no longer proxy the frontend to the same port as the backend when running locally, so locally you must set this variable `REACT_APP_SERVER_URL=http://localhost:3001` for it to work with a backend hosted on port 3001.

We must also add this env var to the GitHub repo to pass into workflows.


frontend/src/util/axios.ts
```diff
@@ -0,0 +1,15 @@
+import axios from 'axios';
+
+const instance = axios.create({
+  baseURL: process.env.REACT_APP_SERVER_URL,
```
pratyush1712, line 4: why don't you simply do `process.env.NODE_ENV == 'production' ?  localhost:3001 : productionURL`?
We can get the production url from the current url using `window.path.split()`

pratyush1712 (commented):
pls also look into why the driver's stats test is failing.

frontend/src/util/axios.ts
Enochen: I run another server that takes up port 3001, so I actually had to run it on a different port and switch the frontend proxy. I'm not too sure about further hardcoding the server url.

pratyush1712 (changes requested):
could you also implement a github action for previews, so that we can automatically set the react_app_server_url [deploy-preview](https://github.com/marketplace/actions/deploy-preview-environment) and [destroy-preview](https://github.com/marketplace/actions/destroy-preview-environment)
Otherwise, I will have to add the server_url to each of the preview environments

pratyush1712 (approved):

pratyush1712 (approved):
Thank you for improving the security of the backend and still keeping it testable. It works both locally and in production!

pratyush1712 (approved):
still works as intended in production. lgtm

PR #442, opened 2023-04-25
Admin-Endpoint Testing

Atikpui007:
### Summary 

This pull request is the first step towards testing the Admin Endpoints

- [x] Tested POST 
- [x] Tested PUT
- [x] Tested GET
- [x] Tested Delete

To run this test suite run command : `npm run test:server`

### Test Plan
Screenshot 2023-04-25 at 16 44 04

### Notes 

- The `put` route was changed because there was no checking for non-existent IDs. If an ID did not exist it will populate the database instead of fail. I changed this to ensure only Admins in the database could be updated. 
- Testing was mainly done on Malformed, Non-Existent and Appropriate data
- Additional Test cases would be added to include more cases of Malformed data and Unauthorized access


harrisonchin2523 (changes requested):
Nice work on these tests Desmond, you were very thorough with the tests and you separated the data into another file which is also a plus. The only thing that I noticed is that when you test the retrieval of all admins, you don't check for what the endpoint returns, just that the res status was 200. Otherwise, once the drivers tests are updated with Enoch's PR, everything else looks good.

server/src/router/admin.ts
```diff
@@ -36,7 +36,12 @@ router.put('/:id', validateUser('Admin'), (req, res) => {
     params: { id },
     body,
   } = req;
-  db.update(res, Admin, { id }, body, tableName);
+
+  const Update = () => {
+    db.update(res, Admin, { id }, body, tableName);
```
pratyush1712, line 41: Thanks for noticing this. This is a problem with dynamoose: https://github.com/dynamoose/dynamoose/issues/259.
If I have to be nitpicky, I would recommend just making the Update an anonymous function, because this function is not used anywhere else, but thanks for resolving this issue!
```
db.getById(res, Admin, id, tableName, (result) => {
  db.update(res, Admin, { id }, body, tableName);
});
```

server/tests/admin-test-data.ts
```diff
@@ -0,0 +1,28 @@
+import express from 'express';
+
+// Basic Data: Appropriate Admin
+export const AdminRequest1 = {
```
pratyush1712, line 4: for now, could you just move all the data into the admin.test.ts file? currently, we don't have separate data files for other testing files. It's good to have a testing data folder, but we can incorporate that once we have all the tests merged in; otherwise, we would lose consistency across all the tests. up to you tho

pratyush1712 (commented):
Please incorporate Harrison's requested changes, it's required to make sure the data returned is correct. Otherwise, looks good to me!

Atikpui007:
Changes suggested by Harrison and Pratyush have been incorporated.

- Added extra test cases to handle edge cases. 
- Unauthorized access, Invalid data 
- Added expect block for `GET` all admins. Compares admins in the table to GET data. 

Changed the put request in `router/admin.ts` to make use of conditionalupdate as referenced in the issue Pratyush linked above. 

###Notes 
- Invalid information sent as requests results in a 500-level error and I believe this is because the data validation is not done before the request is sent to the server so it fails to add to the document since the malformed data does not match the required schema.

server/src/router/admin.ts
```diff
@@ -36,7 +38,10 @@ router.put('/:id', validateUser('Admin'), (req, res) => {
     params: { id },
     body,
   } = req;
-  db.update(res, Admin, { id }, body, tableName);
+
```
pratyush1712, line 41: Looks pretty good! One thing I would recommend is moving this condition to the update function in the `commons.ts` file so that it works for all the usage of updates across all the routes. Then, you won't have to use the conditional update function

PR #443, opened 2023-04-26
Refectored Database to account for admin types

pratyush1712:
## Summary

Basing it off of @Enochen 's auth refactor branch, I have refactored the database to allow the users to specify types of admins: `sds-admin || redrunner-admin`

This pull request also fixes a lot of issues with creating new accounts: driver, admin, and both admin-driver.

## Refactored Database Schema
- The `Admin` model now contains an `isDriver` field that will be set to true if an employee is both a driver and an admin.
- The `Admin` model now also has a `type` field that is a an array which can include the type of admin: `sds-admin` and/or `redrunner-admin`. I think it should be changed to organization instead.

## Revamped Server Routes
- The `Driver Stats` will now send responses wrapped inside a data field for consistency across all the routes.

## Frontend Logic Changes
- If a one selects both `driver` and one or both of `redrunner-admin`/`sds-admin` there will be two requests creating a driver and an admin object with the same id.
- If one want to promote/demote/also make a driver an admin, the frontend has support for that now. Appropriate requests will be send according to this `onSubmit` function:
https://github.com/cornell-dti/carriage-web/blob/12e68f8c74e71689e9ada0fa7bc3c99fa856661b/frontend/src/components/EmployeeModal/EmployeeModal.tsx#L198-L281
- When trying to see a particular employee (admin/driver), instead of using the `useEmployees` function to get all the `drivers` and `admins` and then searching for the said employee, the frontend, will now make a request to the backend for that specific employee (it's easier to do because then we don't have to worry about common types.)
- If an employee is both `admin` and `driver`, their information will be available at the `admins/:id` endpoint on the frontend. It will fetch both the `admin` and the `driver` and combine them.
https://github.com/cornell-dti/carriage-web/blob/12e68f8c74e71689e9ada0fa7bc3c99fa856661b/frontend/src/components/UserDetail/EmployeeDetail.tsx#L201-L234

## Test Plan
- Try adding and editing drivers/admins: new user, promoting a driver to admin, demoting an admin to driver, making a driver admin as well, making an admin driver as well.
- Make sure you can redirect to the said user's profile after all these changes and try refreshing the page.
- Try adding some rides and make sure they are reflected properly in the stats section of the driver.
- Make sure the Analytics page shows correct data.
- Finally, run `npm run test:server` to make sure backend is working properly.


frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -188,9 +176,13 @@ const EmployeeCards = () => {
 
   return (
     <div className={styles.cardsContainer}>
-      {allDrivers &&
-        allDrivers.map((driver) => (
-          <EmployeeCard key={driver.id} id={driver.id} employee={driver} />
+      {filteredEmployees &&
+        filteredEmployees.map((employee) => (
```
Enochen, line 180: This can be simplified with `?.`

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -32,19 +33,22 @@ type EmployeeModalProps = {
 };
 
 type AdminData = {
+  id?: string;
   firstName: any;
   lastName: any;
+  type?: string[];
+  isDriver?: boolean;
```
Enochen, line 40: should not be optional

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -132,57 +145,137 @@ const EmployeeModal = ({
...
+      isDriver: selectedRole.includes('driver'),
+    };
+
+    const existingDriver = existingEmployee?.isDriver === undefined;
+    const existingAdmin = existingEmployee?.isDriver !== undefined;
+
+    if (existingEmployee) {
+      if (selectedRole.includes('driver')) {
+        if (selectedRole.some((role) => role.includes('admin'))) {
+          if (existingDriver && existingAdmin) {
```
Enochen, line 227: I think this `if` chain would be more readable if the driver/admin logic is separated

also, either operations don't need to be sequential, so you can use Promise.all()

Enochen (changes requested):

frontend/src/components/EmployeeModal/EmployeeModal.tsx
Enochen: because of bandwidth limitations, let's tackle this next sem

Enochen (approved):

PR #444, opened 2023-05-01
Implemented Loading Screen

RohitValiveti:
### Summary 

This is a UI change for when data is being fetched. I implemented a loading spinner for when data is being fetched and displayed it appropriately. Prior to this change, when a user's rides were being fetched, the "NoRidesView" was being displayed, even if a user has rides, but the API fetch had not yet finished. With this loading screen, a user will be able to differentiate between if the data is being loaded, if they have rides, or if they actually do not have rides.

This loading screen can also be reused in other areas of the codebase for when data is being loaded.

### Test Plan 

Nothing breaking.

### DIFF
--- a/frontend/src/pages/Rider/Schedule.tsx
+++ b/frontend/src/pages/Rider/Schedule.tsx
@@ -13,6 +13,7 @@ import AuthContext from '../../context/auth';
 import NoRidesView from '../../components/NoRidesView/NoRidesView';
 import RequestRideModal from '../../components/RequestRideModal/RequestRideModal';
 import Notification from '../../components/Notification/Notification';
+import LoadingScreen from '../../components/LoadingScreen/LoadingScreen';
 import styles from './page.module.css';
 
 const Schedule = () => {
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Rider/Schedule.tsx
```diff
@@ -21,13 +22,17 @@ const Schedule = () => {
...
+  const [isLoading, setIsLoading] = useState<boolean>(true);
   const { id, user } = useContext(AuthContext);
   const { withDefaults } = useReq();
   document.title = 'Schedule - Carriage';
   const refreshRides = useCallback(() => {
     fetch(`/api/rides?rider=${id}`, withDefaults())
       .then((res) => res.json())
-      .then(({ data }) => componentMounted.current && setRides([...data]));
+      .then(({ data }) => {
+        componentMounted.current && setRides([...data]);
```
SGupta101, line 33: This line seems to be resulting in an "Unhandled Rejection (TypeError): data is not iterable" error when I open the schedule page as a student.

SGupta101 (changes requested):
Good job on this PR! The change, however, seems to be throwing an error.

PR #445, opened 2023-05-05
making button and tags more descriptive

harrisonchin2523:
### Summary 

We needed to make the edit button more descriptive because it opened a different modal depending on whether the ride was recurring or not. In addition, we wanted to make the dates of the recurring rides more prevalent as well as the tags becoming more descriptive of the location of the pickup and dropoff locations.

### DIFF
--- a/frontend/src/components/RequestRideModal/RequestRideModal.tsx
+++ b/frontend/src/components/RequestRideModal/RequestRideModal.tsx
@@ -71,7 +71,7 @@ const RequestRideModal = ({
               : () => openCreateOrEditModal('EDIT_REGULAR')
           }
         >
-          Edit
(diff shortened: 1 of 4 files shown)


Atikpui007:
Looks good. 
Repeating dates are much clearer with the added highlights. 
The buttons and the modals are also more descriptive. I tested this on different browsers and it is also pretty responsive so that's good

A concern
The highlighted dates are right next to the days which makes it a bit harder to read. If you could space that out a bit that would be great

SGupta101 (changes requested):
Good job on this PR! I think the description and spacing looks good. The only concern I had is that when I check the "repeating" check-box, the outline seems to be longer than the square of the check-box.

PR #446, opened 2023-05-13
Sg2265/carriage-login

SGupta101:
### Summary 

Started modifying carriage-login based on new Figma design


CollinWoo:
Great work on this feature Stuti! The footer looks good for the most part but the images in the footer don't seem to show up on my end. I was able to fix this by importing the images explicitly at the top of the file (e.g. `import group525 from './Group525.png'`) and then adding this import as the source of the image (e.g. `</img>`) instead of hard-coding the path directly. Otherwise, everything looks great!

frontend/src/pages/Landing/Group525.png
CollinWoo, line 1: I think the images should be renamed to be more descriptive. "Group525" doesn't really convey what this image is and may be confusing when referring back to the code in the future.

frontend/src/pages/Landing/Group526.png
CollinWoo, line 1: Same concern as with the other picture.

frontend/src/pages/Landing/Footer.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { ReactElement } from 'react';
+import styles from './landing.module.css';
+import { dti_logo } from '../../icons/other';
+
```
CollinWoo, line 4: I think there should be an explicit import for the images used in the footers here (e.g. `import group525 from './Group525.png'`) since the images were not showing up for me otherwise and it makes the code cleaner by avoiding hard-coding paths in the body of the code.

frontend/src/pages/Landing/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+            paddingTop: 30,
+            paddingBottom: 30,
+          }}
+        >
+          Students schedule, <br />
+          edit, and cancel rides <br />
+          on rider web <br />
+          and mobile.
+        </p>
+        <img src="frontend/src/pages/Landing/Group525.png" />
```
CollinWoo, line 24: I think this should be changed to have the src use the imported image from the suggestion above.

frontend/src/pages/Landing/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+            paddingTop: 30,
+            paddingBottom: 30,
+          }}
+        >
+          Administrators add <br />
+          students and <br />
+          employees and <br />
+          dispatch rides.
+        </p>
+        <img src="frontend/src/pages/Landing/Group525.png" />
```
CollinWoo, line 43: I think this should be changed to have the src use the imported image from the suggestion above.

CollinWoo (commented):
Great work on this feature Stuti! The footer looks good for the most part but the images don't seem to show up on my end. I was able to fix this by importing the images explicitly at the top of the file (e.g. import group525 from './Group525.png') and then adding this import as the source of the image (e.g. </img>) instead of hard-coding the path directly. Otherwise, everything looks great!

PR #456, opened 2023-09-18
Fix Deployment Issues

pratyush1712:
### Summary 

This pull request fixes the env variable that is used while deploying. `OAUTH_CLIENT_ID` is the same as `REACT_APP_CLIENT_ID`, but one is for the server and the other for the frontend.


Atikpui007 (approved):
Deployment issue fix👍🏽

PR #462, opened 2023-09-29
Add Schedule Events selectable and interactable via keyboard

Enochen:
### Summary 

Fixes Issue 3 of [Keyboard Testing task](https://trello.com/c/iNd610Tg/486-keyboard-testing)

### Test Plan 

Within the homepage, you can now Tab onto each event on the schedule and press "Enter" or "Space" to toggle the popup modal for the selected event.

### Notes 

Implementation was kind of hacky but it was the only way to do it because the element we needed to add the `tabIndex` property to was part of the `react-big-calendar` library which limits exposure to its internal components.

### Breaking Changes  

I had to update the `react-big-calendar` library we use from `^0.33.5` to `^1.8.4`. This introduced minor styling shifts (for example the text align of the left column of the schedule). One should compare this build to current staging to see what changed via the new library version.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -30,7 +30,7 @@
         "jwt-decode": "^3.1.2",
         "moment": "^2.29.4",
         "react": "^17.0.1",
-        "react-big-calendar": "^0.33.5",
+        "react-big-calendar": "^1.8.4",
         "react-csv": "^2.0.3",
         "react-datepicker": "^3.8.0",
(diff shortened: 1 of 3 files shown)


CollinWoo:
Great work Enoch! I tested out the keyboard shortcut myself and it worked perfectly. I like how simple and clean the code you wrote is. I think this feature will be great for accessibility and easily navigating the homepage in the future.

raissaji (approved):

PR #463, opened 2023-09-29
Refocus calendar button after calendar modal close in Home page 

Enochen:
### Summary 

Fixes Issue 4 of [Keyboard Testing task](https://trello.com/c/iNd610Tg/486-keyboard-testing)

### Test Plan 

After opening the calendar and selecting +interacting with a day (thus closing the calendar popup), browser focus is now pointed at the button that opens the calendar, instead of an invisible element like before.

### Notes 

I had to do a weird trick with the `customInputRef` prop in the `DatePicker` component to get the ref to the button to persist. Unfortunately, `react-datepicker` doesn't seem to provide a better way to focus on our custom button input (at least I couldn't find anything).

### DIFF
--- a/frontend/src/components/MiniCal/MiniCal.tsx
+++ b/frontend/src/components/MiniCal/MiniCal.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import DatePicker from 'react-datepicker';
+import React, { useRef, forwardRef, HTMLProps } from 'react';
+import DatePicker, { ReactDatePickerProps } from 'react-datepicker';
 import cn from 'classnames';
 import 'react-datepicker/dist/react-datepicker.css';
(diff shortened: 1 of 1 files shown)


Atikpui007 (changes requested):
Very thorough PR

I did some research and could not find anything on native react-date picker methods that could address the issue like you said. 

Good use of the custom Input Ref to address the focus issue 

I performed the test and it works as expected 
Some notes - I know this wasn't part of the ticket but since there are buttons without ARIA labels can you ensure they are correctly labeled so as not to come back to make accessibility changes

namanhboi:
I pulled the PR and after selecting a day, it indeed refocuses on the calendar element. PR works as expected.

PR #465, opened 2023-10-02
Add alt text to back button

CollinWoo:
### Summary 

This pull request adds alt text to the chevron button that takes users from the employee details page back to the employees list page.

image

- [x] Added alt text to the chevron image for the back button
- [x] Documented the compRides() function 

### Test Plan 
Testing was conducted by using inspect element on the back button on the employee details webpage and verifying that the image has the `alt` attribute with the appropriate text:

Screenshot 2023-10-01 at 11 28 16 PM

### DIFF
--- a/frontend/src/components/UserDetail/EmployeeDetail.tsx
+++ b/frontend/src/components/UserDetail/EmployeeDetail.tsx
@@ -145,7 +145,7 @@ const Header = () => {
         }}
         className={styles.header}
       >
-        <img className={styles.chevronLeft} src={chevronLeft} />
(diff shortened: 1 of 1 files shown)


pratyush1712 (changes requested):
thank you for adding documentation! could you change the alt text from "Employees" to "Back to Employees List" because that makes more sense for the back buttton here. also, could you also make sure all the back buttons on the site has alt text (i think there might be more than just the employees button)

pratyush1712 (approved):

PR #471, opened 2023-10-22
Calendar Modal Fix - Multiple Selections Display

Atikpui007:
### Summary 

This pull request fixes the bug displayed on the Mini Calendar when selecting dates using the Today and Tomorrow buttons. Also addresses closing the modal when a date is selected (ideally not the intended behavior)

- [x] implemented useRef state to address the change to the DatePicker component
- [x] fixed closing modal when date is selected 

### Test Plan 

Choose dates using the Today and Tomorrow buttons when other dates are selected to identify behavior

#### Initial behavior
https://github.com/cornell-dti/carriage-web/assets/106533812/42d24d92-6707-4826-a15d-0d1a078a475f

#### Fixed behavior
https://github.com/cornell-dti/carriage-web/assets/106533812/456a4b64-319a-497e-acf6-734a78e01117

### Breaking Changes  
None

### DIFF
--- a/frontend/src/components/MiniCal/MiniCal.tsx
+++ b/frontend/src/components/MiniCal/MiniCal.tsx
@@ -33,6 +33,7 @@ const Icon = () => (
 
 const MiniCal = () => {
   const { curDate, setCurDate } = useDate();
+  const datePickerRef = React.useRef<any>(null);
 
   const updateDate = (d: Date) => {
     setCurDate(d);
@@ -82,7 +83,9 @@ const MiniCal = () => {
     <div className={styles.root}>
       <DatePicker
         adjustDateOnChange
+        ref={datePickerRef}
         selected={curDate}
+        shouldCloseOnSelect={false}
         onChange={updateDate}
         closeOnScroll={true}
         dateFormat="MMM dd, yyyy"
@@ -103,7 +106,11 @@ const MiniCal = () => {
               <button
                 className={cn(styles.btn2, { [styles.active]: isToday(date) })}
                 onClick={() => {
-                  updateDate(new Date());
+                  const today = new Date();
+                  if (datePickerRef.current) {
+                    datePickerRef.current.setSelected(today);
+                  }
+                  updateDate(today);
                   pseudoScroll();
                 }}
               >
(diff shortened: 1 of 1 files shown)


frontend/src/components/MiniCal/MiniCal.tsx
```diff
@@ -33,6 +33,7 @@ const Icon = () => (
 
 const MiniCal = () => {
   const { curDate, setCurDate } = useDate();
+  const datePickerRef = React.useRef<any>(null);
```
Enochen, line 36: We can extend the types via

```
declare module 'react-datepicker' {
  export default class ReactDatePicker {
    setPreSelection?: (date: Date) => void;
  }
}
```

to be able to do

```suggestion
  const datePickerRef = React.useRef<DatePicker>(null);
```

Enochen: imo we don't want to introduce more `any`s into the system if we don't have to

frontend/src/components/MiniCal/MiniCal.tsx
```diff
@@ -103,7 +106,11 @@ const MiniCal = () => {
               <button
                 className={cn(styles.btn2, { [styles.active]: isToday(date) })}
                 onClick={() => {
-                  updateDate(new Date());
+                  const today = new Date();
+                  if (datePickerRef.current) {
+                    datePickerRef.current.setSelected(today);
+                  }
```
Enochen, line 112: Optional chaining works with function calls too; also `setPreSelection` seems more appropriate than `setSelected` here (likely less overhead). It's also what's used in [`react-datepicker`'s native implementation of the "Today" button](https://github.com/Hacker0x01/react-datepicker/blob/main/src/calendar.jsx#L731).

```suggestion
                  datePickerRef.current?.setPreSelection?.(today);
```

Enochen (approved):
Works on my machine but have few suggestions

SGupta101 (commented):
Great work on this PR Desmond! I think the changes look good, I tested them on different browsers and it seems to work properly. Because the CustomInput function does not have an internal state, it can be converted to a functional component to make the code a bit shorter, but it is not needed.

PR #472, opened 2023-10-23
Sg2265/color-contrast

SGupta101:
Fixing color contrast issues across site.
1. Logout:
2. NetIDs everywhere:
3. Employees - add OR edit an employee modal:
4. Employees - card - delete:
5. Analytics:
6. Home - Select Dates Modal:
7. Error messages:


frontend/src/components/EmployeeModal/EmployeeInfo.tsx
```diff
@@ -11,6 +11,16 @@ type EmployeeInfoProps = {
   phone?: string;
 };
 
+// The component accepts firstName, lastName, netId, and phone as optional props,
```
CollinWoo, line 14: Nice documentation! But I think that instead of using single line comments `//`, you should wrap this documentation block in a JSDoc comment using `/** */` since this is the standard way of writing documentation in Javascript.

Atikpui007:
merging old PR which works on all browsers

Atikpui007 (approved):

PR #478, opened 2023-10-31
Display employee photos in cards

Enochen:
### Summary 

This allows photos to be visible. The issue was that admin accounts supported photos but no admin account in the db actually has a photo. However the corresponding driver accounts do have photos.

This PR "combines" the two employee account objects so that you only need a `photoLink` in either your admin or driver account for it to display in your card.

Note that this will use the driver `photoLink` if both are provided. In the future we should aim to restructure the DB so that this kind of ambiguity isn't possible.

### Test Plan 
image

### Notes 

Note that this will use the driver `photoLink` if both are provided. In the future we should aim to restructure the DB so that this kind of ambiguity isn't possible.

### DIFF
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -156,27 +156,26 @@ const EmployeeCard = ({
 const EmployeeCards = () => {
   const { admins, drivers } = useEmployees();
 
-  const allEmployees = [...admins, ...drivers];
-  const adminIds = new Set(admins.map((admin) => admin.id));
-  const filteredEmployees = allEmployees.filter((employee: Employee) => {
-    // if not admin (means driver), check if another admin is representing this driver
-    if (employee['isDriver'] == undefined) return !adminIds.has(employee.id);
-    return true;
+  const employees: Record<string, DriverType | AdminType> = {};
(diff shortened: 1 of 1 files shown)


frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -156,27 +156,26 @@ const EmployeeCard = ({
...
 
-  const allEmployees = [...admins, ...drivers];
-  const adminIds = new Set(admins.map((admin) => admin.id));
-  const filteredEmployees = allEmployees.filter((employee: Employee) => {
-    // if not admin (means driver), check if another admin is representing this driver
-    if (employee['isDriver'] == undefined) return !adminIds.has(employee.id);
-    return true;
+  const employees: Record<string, DriverType | AdminType> = {};
+  [...admins, ...drivers].forEach((employee) => {
+    employees[employee.id] = { ...employees[employee.id], ...employee };
```
CollinWoo, line 161: I think one potential concern with this array destructuring is that fields other than `photolink` could be overridden as well if they differ between the driver and the admin. I'm not sure how often this would occur though.

CollinWoo (commented):
Great work Enoch! The employee photos now show up perfectly on my end. This seems like a good fix for now, but I definitely agree that more consideration should be taken with overriding fields between the Admin and Driver classes

PR #485, opened 2023-11-29
implement employee filters (admins/drivers)

Enochen:
### Summary 

This diff adds a two-button role filter to the Employees page.

For example, selecting "Drivers" will filter the results such that only Driver accounts are shown. Selecting both shows all employees, as does selecting none (the default state).

### Test Plan 

<details><summary>None Selected</summary>

img

</details>
<details><summary>One Selected</summary>

img

</details>
<details><summary>Both Selected</summary>

img

</details>

### Notes 

The described button behavior is pretty arbitrary imo, and can be changed in the future.


kevin-lin12:
Looks great! The styling is on point with the other buttons, and the function works perfectly fine.

The only concern is how a user would want to filter to list out only users with BOTH Admin and Driver tags. If both are selected, it shows all the users

SGupta101 (approved):
Great job on this PR Enoch! Tested on multiple browsers and works as expected.

PR #486, opened 2023-11-29
Standardize how document title is set

Enochen:
### Summary 

Previously we were setting document titles in inconsistent ways across pages.

This diff makes them all consistent, with useEffect to prevent unnecessary title updates.
 
### Test Plan 

Made sure all the titles stayed the same.

### DIFF
--- a/frontend/src/pages/Admin/Analytics.tsx
+++ b/frontend/src/pages/Admin/Analytics.tsx
@@ -19,7 +19,7 @@ const Analytics = () => {
 
   React.useEffect(() => {
     document.title = 'Analytics - Carriage';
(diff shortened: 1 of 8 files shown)


pratyush1712 (approved):
I am not completely sure if using `useEffect` is better than setting the document title in the function itself. However, it works perfectly fine and `useEffect` makes it easier to understand the code. Awesome work!!

Enochen:
The advantage of it being consistent + the flexibility of being able to make currently static titles potentially reactive to state in the future was the main goal. I think performance-wise there shouldn't be an observable difference either way.

PR #487, opened 2023-12-03
Employee Modal Update Image

pratyush1712:
### Summary

This pull request is the first step towards improving the user experience when editing or creating an employee. While working on the task, I found about a bug where editing an employee's image was not working properly because of some inconsistency. I resolved all the issues with editing employee images.

Along with that, I also made sure the images were updated on the client side without the need to refresh.
**Issue**: When updating the user image, the URL remains the same. Therefore, react does not refresh the image.
**Solution**: I changed the URL by adding the current time. That way the image URL is always different and react will render the right image after an employee has been updated without needing to refresh.

I also improved the code quality and added documentation to one of the most complicated functionalities on the project: adding and editing an employee. Because the same modal is used to edit, create, promote, demote, and delete an employee (admin or driver), the component's logic is highly complicated. I broke down and combined some functions to improve readability as well as added JSDocs.

### Test Plan

- Run all possible combinations: editing, creating, promoting, demoting, and deleting employees.
- Try creating a new employee with an image and then try to edit the image.


frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -128,7 +128,7 @@ const EmployeeCard = ({
         firstName={firstName}
         lastName={lastName}
         netId={netId}
-        photoLink={photoLink}
+        photoLink={photoLink ? photoLink + '?v=' + new Date().getTime() : ''}
```
Enochen, line 131: tiny thing but I think using string interpolation could make the string manipulation clearer

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -81,6 +81,14 @@ const EmployeeModal = ({
     setIsOpen(false);
   };
 
+  /**
+   * The function `parseAvailability` takes an array of objects containing `startTime`, `endTime`, and
+   * `days` properties, and returns an object where each day is a key and the value is an object
+   * containing the `startTime` and `endTime` for that day.
+   * @param {ObjectType[]} availability - The `availability` parameter is an array of objects. Each
+   * object in the array represents a time slot with the following properties:
```
Enochen, line 89: I think this line is cut off short (I don't see any following properties)

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -225,57 +296,141 @@ const EmployeeModal = ({
       if (selectedRole.includes('driver')) {
         if (selectedRole.some((role) => role.includes('admin'))) {
           if (existingDriver && existingAdmin) {
-            await createOrUpdateDriver(driver, false);
-            await createOrUpdateAdmin(admin, false);
+            await createOrUpdateEmployee(
```
Enochen, line 299: This is a lot of calls to `createOrUpdateEmployee` with basically two sets of params. One way to make this more concise imo is to create two helper functions for drivers and admins. It looks like the last three params are always the same for each scenario

frontend/src/components/UserDetail/EmployeeDetail.tsx
```diff
@@ -296,7 +296,11 @@ const EmployeeDetail = () => {
             netId={employee.netId}
             employee={employee}
             role={roleValue()}
-            photoLink={employee.photoLink}
+            photoLink={
+              employee.photoLink
+                ? employee.photoLink + '?v=' + new Date().getTime()
```
Enochen, line 301: same comment as above - also this totally could go into a helper function because it's very specific behavior that if updated needs to be updated everywhere

Enochen (changes requested):
looks good, tho i'm unsure about why `package-lock.json` was modified if this diff doesn't include a change to `package.json`

namanhboi:
I pulled and tested this PR and the Employee image does load immediate without having to refresh. Great PR, works as intended

PR #493, opened 2024-02-26
Rename accessibility enums

andrewschoi:
We update `accessibility` to include `None` and `tag` to include `east`.


Atikpui007 (approved):
Simple PR

PR #495, opened 2024-03-02
Allow phone number dashes

selenaliu1:
### Summary 

- edited Rider modal and Employee modal to allow dashes in the phone number

### Test Plan 

Valid inputs:
Screenshot 2024-03-02 at 11 49 40 AM
Screenshot 2024-03-02 at 11 59 48 AM
Invalid inputs (too long):
Screenshot 2024-03-02 at 12 00 05 PM

### Notes 

- Accepts 10-12 characters, but only takes the first 10 numbers

### Breaking Changes  

### DIFF
--- a/frontend/src/components/EmployeeModal/EmployeeInfo.tsx
+++ b/frontend/src/components/EmployeeModal/EmployeeInfo.tsx
@@ -75,13 +75,13 @@ const EmployeeInfo = ({
           type="tel"
           defaultValue={phone}
           min={10}
-          max={10}
+          max={12}
           aria-required="true"
           className={cn(styles.input)}
           ref={register({
             required: true,
-            pattern: /[0-9]{10}/,
-            maxLength: 10,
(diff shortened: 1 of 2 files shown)


raissaji (commented):
Looks good! A suggestion is to explicitly define where the dashes should be, so that a phone number like 40-98204-2 will be flagged as invalid.

andrewschoi:
`maxLength` and `minLength` are not needed. Change `pattern` to be a regular expression for a phone number.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions

```
            ref={register({
              required: true,
              pattern: /^[0-9-]{10,}$/,
              maxLength: 12,
              minLength: 10,
            })}
```

Atikpui007 (changes requested):
Define a validity check to ensure that dashes cannot just be placed anywhere.

PR #496, opened 2024-03-03
Unscheduled rides are now expanded by default

raissaji:
### Summary 

This pull request ensures that the "Unscheduled Rides" dropdown is expanded by default.

### Test Plan 

Screen Shot 2024-03-03 at 12 06 27 AM

### Notes 

The "Unscheduled Rides" dropdown for every date are not independent states. For example, if I un-expand the dropdown for March 21th, and then click on March 4th, the latter's dropdown will be unexpanded instead of defaulting to its expanded state.

Screen Shot 2024-03-03 at 12 05 55 AM

*I unexpand "Unscheduled Rides" on March 21*

Screen Shot 2024-03-03 at 12 06 03 AM

*When I go back to March 4, it remains unexpanded*

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -37,7 +37,7 @@
     "typescript": "^4.8.4"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "react-scripts --openssl-legacy-provider start",
     "build": "react-scripts build",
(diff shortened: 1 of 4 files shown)


namanhboi:
Short and simple changes, looks good to me!

Atikpui007 (approved):
Intuitive changes looks good

PR #497, opened 2024-03-03
Add Needs entry to student when selecting "other"

namanhboi:
### Summary

Added an input field that appears below the "Needs" drop downbox when "Other" is selected to student. It doesn't update the database and is for just appearance.

image
_When "Other" is not selected_

image
_When "Other" is selected_

Tested on different screen sizes (mobile, desktop) and no bugs have been spotted.

- [x] implemented the textbox when other is selected.

### Test Plan

No plans currently as it doesn't update the database or have any bugs.

### Breaking Changes  

None.

### DIFF
--- a/frontend/src/components/Modal/RiderModalInfo.tsx
+++ b/frontend/src/components/Modal/RiderModalInfo.tsx
@@ -153,7 +153,7 @@ const RiderModalInfo = ({
               id="otherNeeds"
               name="otherNeeds"
               type="text"
-              aria-required="true"
+              placeholder='Please Specify Needs'
             ></Input>
           ) : (
             <></>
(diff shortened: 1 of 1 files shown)


selenaliu1 (commented):
Overall this looks good to me. My only suggestion would be to add a label like "Please Specify" so that it is very clear what the input field is for.

namanhboi:
> Overall this looks good to me. My only suggestion would be to add a label like "Please Specify" so that it is very clear what the input field is for.

I just added the placeholder text, thank you for the feedback!

Atikpui007 (approved):
Simple changes looks good 
Will need this for subsequent PR changes

PR #499, opened 2024-03-04
Temp deploy fix [DO NOT MERGE]

Atikpui007:
### Summary
Implemented the `--openssl-legacy-provider` flag in our `start` and  `build` script to address an OpenSSL-related error encountered with newer versions of Node. This change allows our application to bypass the `ERR_OSSL_EVP_UNSUPPORTED` error, enabling compatibility with older dependencies that require legacy OpenSSL support.

### Notes
Encountered a runtime error due to OpenSSL version incompatibility. The issue arises from the stricter default settings for OpenSSL in Node.js v17 and above, affecting older dependencies:

A link to a similar issue with the outdated cryptographic dependency here 
[Issue](https://github.com/nodejs/help/issues/4115)

```plaintext
Error: error:0308010C:digital envelope routines::unsupported
...
code: 'ERR_OSSL_EVP_UNSUPPORTED'

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,15 +28,9 @@ The **Rider App** branch can be found [here](https://github.com/cornell-dti/carr
 
 **Current Contributors**
 
-- Harrison Chin - Software Developer
-- Andrew Choi - Product Manager
-- Rohit Valiveti - Software Developer
-- Pratyush Sudhakar - Technical Project Manager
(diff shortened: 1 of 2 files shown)


Atikpui007:
Fixed the migration issues and the problems with OpenSSL due to the deprecated cryptojs library

PR #500, opened 2024-03-09
Edit visual indicator for active vs inactive students

selenaliu1:
### Summary 

Edited visual indicator so that red line shows only when a user's status is inactive and their end date has passed.

### Test Plan 

Screenshot 2024-03-09 at 4 01 26 PM
Red lines only when they're inactive.

### Notes 

- Some students with end dates that have passed have activity status active when it should be inactive

### Breaking Changes  


Atikpui007:
Closing this branch because we are moving towards a new implementation for the inactive and active studets

PR #501, opened 2024-03-13
Rj353/student address requirements

raissaji:
### Summary 

This pull request removes the address parsing condition within isAddress. It also removes the requirement that the input be at least 3 characters long in RiderModalInfo.tsx. Now, any address input that is not empty or spaces will be valid. This is to resolve the inconsistent add/edit behavior noted in PR #425 regarding the address field.

### Test Plan 

Screen Shot 2024-03-12 at 9 34 10 PM

Attempting to submit an address input of just spaces will alert the error message.

Screen Shot 2024-03-12 at 9 34 28 PM

Attempting to submit an empty address input will also alert the error message.

Screen Shot 2024-03-12 at 9 35 19 PM

Previously, the input "300 Day Hall" wouldn't have worked. Now, 300 Day Hall is parsed and appended with ", Ithaca, NY 14850" to be a valid address.

### Breaking Changes
None

### DIFF
--- a/frontend/src/components/Modal/RiderModalInfo.tsx
+++ b/frontend/src/components/Modal/RiderModalInfo.tsx
@@ -165,7 +165,7 @@ const RiderModalInfo = ({
             aria-required="true"
             ref={register({
               required: true,
-              pattern: /^[a-zA-Z0-9\s,.'-]{3,}$/,
+              validate: (value) => value.trim() !== '',
             })}
           />
           {errors.address && (
--- a/server/src/util/index.ts
+++ b/server/src/util/index.ts
@@ -11,28 +11,6 @@ export function createKeys(property: string, values: string[]) {
 }
 
 export function isAddress(address: string) {
-  // let parsedAddr;
-  // try {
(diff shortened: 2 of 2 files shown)


frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -165,7 +165,7 @@ const RiderModalInfo = ({
             aria-required="true"
             ref={register({
               required: true,
-              pattern: /^[a-zA-Z0-9\s,.'-]{3,}$/,
+              validate: (value) => value.trim() !== '',
```
namanhboi, line 168: Hi Raissa, you did a great job on this PR as your code was both simple and correct. Just a sidenote, but if you wanted to do this with regex, you could use: `pattern : /^(?!\s*$).+/`, which represents strings with at least one non empty character.

Atikpui007 (changes requested):
Address Nahms comments and also extend this to ensure we still maintain some form of address validation logic.

PR #502, opened 2024-03-15
swap datalist to select for dropoff locations

andrewschoi:


frontend/src/components/RideModal/Pages/RiderInfo.tsx
```diff
@@ -93,34 +93,35 @@ const RiderInfoPage = ({ formData, onBack, onSubmit }: ModalPageProps) => {
           </datalist>
         </div>
         <div className={styles.dropoffLocation}>
-          <Label htmlFor={'dropoffLoc'} className={styles.label}>
+          <Label htmlFor="dropoffLoc" className={styles.label}>
             Dropoff Location
           </Label>
-          <Input
+          <select
```
Atikpui007, line 99: The dropoff location should still be kept as an input field because we intend to type

We can introduce a fuzzy search behavior to allow for the list to be filtered when we search for locations

Atikpui007 (changes requested):
Both the pickup and dropoff locations should have the searching and rendering of lists implemented

PR #503, opened 2024-03-17
Add ability to select multiple needs for the Add / Edit Student Modal

namanhboi:
### Summary 

I added a new React component with styling called Dropdownbox. It is a dropdown checklist of items, allowing for multiple selections, with a confirm button. I also used this component in the Add / Edit Student Modal to enable the selection of multiple needs. This is purely frontend and I haven't made this multiple selection work with the backend.

image
_When the box is not open_
image
_The box is opened and some items are selected_
image
_The box is closed_

I still need to make the textbox for when "Other" is selected work with this new Dropdownbox component. My previous pull request (https://github.com/cornell-dti/carriage-web/pull/497) which implemented the textbox for "Other" was done under the assumption that the select react tag was used. I also need to discuss with Lazim about possible redesigns for the needs selection because I have a different opinion on where the textbox to enter an "Other" need should be placed.  

This pull request is a part of redesigning the Add / Edit Student Modal and specifically the needs selection.

- [x] Implemented multiple selection for needs in the Add / Edit Student Modal
- [x] Created a new React component for dropdown checklists.
- [ ] Need to make the textbox for when "Other" is selected work with this component. 

### Test Plan 

I tested the design on different display sizes and there seems to be an issue where the "Needs: " text isn't in bold while all the other ones are. 

image
_On the Galaxy S20 inspect element_
image
_On the Responsive size in inspect element_

Although none of the text are in bold when i'm just on the website and not in Inspect.

image
_When I'm just on the website_

I will try to figure out this issue in the next PRs.

#### The "Other" needs textbox

In the future when I incorprate the textbox for when "Other" is selected, I need to discuss with Lazim about possible redesigns. This is because implementing the textbox inside the Dropdownbox component itself will make the component not as reusable. I think that the textbox should be moved out of the dropdown checklist and be placed below it.

image
_Lazim's design_

image
_My idea_

#### Styling of the component

I am not very good with css, so a lot of my styling is working directly with px values which might not be good.

#### Behavior when click outside of the dropdown box.

I haven't implemented this later on I will make it so that the box collapses. 

### Breaking Changes

I haven't found any so far.


frontend/src/components/DropdownBox/DropdownBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <div className="dropdown-header-contents">
+          {placeHolderText}
+          {'      '}
+          {isDropdownBoxOpen ? '⇈' : '⇊'}
+        </div>
+      </button>
+
+      {isDropdownBoxOpen && (
+        <>
+          <div className="dropdown-list">
```
Atikpui007, line 50: I noticed the dropdown box was custom styled and this is okay if there is a significant change from the native HTML styling but since they are very similar (essentially the same) lets stick to using native HTML elements as they are designed to be accessible and cross platform 

A suggestion is to map out everything and use `input type = checkbox`

Example 
```
      {isDropdownBoxOpen && (
        <div className="dropdown-list">
          {dataSource.map((itemName) => (
            <label key={itemName} className="dropdown-list-item">
              <input
                type="checkbox"
                checked={selectedItems.has(itemName)}
                onChange={() => toggleSelectItem(itemName)}
              />
              <span className="item-name">{itemName}</span>
            </label>
          ))}
```

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -131,22 +132,15 @@ const RiderModalInfo = ({
...
-                  {value}
-                </option>
-              );
-            })}
-          </select>
+          Needs:
+          <DropdownBox
+            placeHolderText="Select Your Needs"
+            dataSource={Object.values(Accessibility)
+              .splice(1)
```
Atikpui007, line 139: instead of using the splice we can just remove the None value from the Accessibility enum since the stakeholders suggested there is no need for it to be there

Atikpui007 (changes requested):
Overall a pretty good PR and hope you are enjoying working with Typescript. 
Remember to add the conditional render for the text input as well 
Note that it is rendered when `other` is selected so we can think of how to implement that in a generic way to support different uses of it

frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -131,22 +132,15 @@ const RiderModalInfo = ({
...
-          >
-            {Object.values(Accessibility).map((value, index) => {
-              return (
-                <option key={index} value={value}>
-                  {value}
-                </option>
-              );
-            })}
-          </select>
+          Needs:
```
Atikpui007, line 135: The reason why this is not bolded when the screen size is changed is because it is not wrapped using a label. 

The headers for each input type uses a label and there is a media-query which keeps the label bolded on different screen sizes. Check out the code in `src/components/modal/ridermodal.module.css`

Needs should be wrapped like the other labels 
```
          <Label className={styles.label} htmlFor="needs">
            Needs:{' '}
          </Label>
```

namanhboi:
### Summary of changes: 

Added the option to include a text input box in the MultiselectBox component. You can specify the option that you want to be associated with the multiselectbox and when that option is selected, the text input box will appear. You can also specify the placeholder text for the text input box. 
image


Some styling changes: 
- The MultiselectBox now uses default html checkbox styling instead of custom styling. 
- Wrapped the "Needs:" text around a Label, so now it is bold when the screen size changes.
- Changed px to rem in the css of the MultiselectBox component.

image
_This is what the header for the MultiselectBox looks like. Now, uses the correct arrow svg icons from figma_

image
_This is what it looks like when you open the MultiselectBox_

image
_This is what it looks like when you select some elements_

image
_When you select the specified option, in this case "Other", the textbox will appear below with the placeholder text given_

image
_The "Needs" text is now bold when you change screen sizes._

This pull request is a part of redesigning the Add / Edit Student Modal and specifically the needs selection.

- [x] Implemented multiple selection for needs in the Add / Edit Student Modal
- [x] Created a new React component for multiselect box.
- [x] Make the textbox for when "Other" is selected work with this component.

### Test plan

I have tested it in different screen sizes and found that it seems to scale fine on smaller screen sizes. This seems to be the case for the overall website. I will try to add media-query to the css file over the next few commits on the Add /Edit Student Modal.

image

image



### Notes: 
I haven't done the backend for this yet, but will start it soon. 

### Breaking changes

So far, I haven't noticed any yet, but I need to do more extensive testing. They will probably show up when I do the backend for the component.

Atikpui007:
Closing because this requires lots of backend setup before frontend can be implemented

PR #504, opened 2024-03-23
WIP: Deactivated riders should not be assigned to rides

selenaliu1:
### Summary 

Worked on admin side, filtering out the deactivated riders when they add a ride so that only active riders are displayed.

### Test Plan 
Tested through trying to add rider on admin side.

### Notes 

- Still need to work on student side, do not let deactivated students request a ride.
- Should test if admin can add a deactivated student when editing a ride.

### Breaking Changes  

### DIFF
--- a/frontend/src/components/UserTables/StudentsTable.tsx
+++ b/frontend/src/components/UserTables/StudentsTable.tsx
@@ -129,7 +129,7 @@ const StudentsTable = ({ searchName }: studentTableProps) => {
             endDate
           )}`;
           const usageData = getUsageData(id);
-          const isStudentInvalid = moment().isAfter(moment(endDate)) && active;
+          const isStudentInvalid = moment().isAfter(moment(endDate)) && !active;
           const location = {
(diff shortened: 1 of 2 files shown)


frontend/src/components/UserTables/StudentsTable.tsx
```diff
@@ -129,7 +129,7 @@ const StudentsTable = ({ searchName }: studentTableProps) => {
             endDate
           )}`;
           const usageData = getUsageData(id);
-          const isStudentInvalid = moment().isAfter(moment(endDate)) && active;
+          const isStudentInvalid = moment().isAfter(moment(endDate)) && !active;
```
raissaji, line 132: I think this should be || instead of &&.

raissaji (commented):
Good work! WIP is simple so far, so no substantial improvements from me. However, something to consider is how you want to update/re-retrieve the list of valid rider IDs (lines 67-79) when you activate a deactivated rider.

PR #507, opened 2024-03-25
Select Component - dropdown with fuzzy search and multi-select capability

Atikpui007:
### Summary 

This pull request is aimed at introducing a new component / library which solves the problem of having searchable select dropdowns

- [x] installed `react-select` library which comes with predefined features such as multi-select, searchable select dropdown, etc
- [x] implemented form functionality for the Select dropdown and can correctly pass information to the backend
- [x] implemented rules for the form functionality to ensure only valid data is sent through
- [x] Added more custom functionality to the select component (returns only the value field to the form now)
- [x] Replaced select dropdowns for Ride modal
- [x] Replaced select dropdowns for Student modal 
- [x] Replaced select dropdowns for other modals 

The goal is to ensure that all the dropdowns make use of this component 

### Test Plan 
To test the functionality of `react-select` 

1. Open the Add / Edit Ride modal
2. Under the Rider Info section the pickup and dropoff location make use of React-Select 

https://github.com/cornell-dti/carriage-web/assets/106533812/391f211c-e0fa-4924-9d2f-2df9d73b364f

### Notes 
Read on the [react-select](https://www.npmjs.com/package/react-select) component here

`React-select` library requires a specific type of options to work 
```
options = {
value
label
}

```
~~So it is necessary to transform the datalist into a supportable type for the Select component ~~
Fixed the react-select component to return only the value

### Breaking Changes  
~~Bug with the transforming of options - results in multiple options having the same keys. Causes the website to build a map that is arbitrarily large.~~
Squashed bug by introducing dependency array into the useEffect hook in the `riderInfo.tsx` file. Issue was with the constant updating of the dropdown menu causing performance and web crashes


frontend/src/components/FormElements/FormElements.tsx
```diff
@@ -73,3 +74,32 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     );
   }
 );
+
+type DataList = {
+  id: string;
+  name: string;
+};
```
namanhboi, line 82: The library looks really promising and seems to fulfill almost all of our use cases. One thing I was wondering is should the type for the datalist be a list of objects (Datalist) with property id and name. I think that a lot of what we will be representing as options in a dropdown list will be just lists of string, often from an enum class like Accessibility in types/index.ts.

Atikpui007: The datalist is a requirement for how the react-select library works 

It requires a `label` and a `name` and I noticed that a lot of the things have both an `id` and `name` so it made sense group them like that

namanhboi:
Great work on incorporating the react-select library. It works as expected and allows for easy multi select dropdown boxes. However, now we have to fix some styling issues: 
Cheese_Sat-05Oct24_11 25

Cheese_Sat-05Oct24_11 25_000
_Here, the dropdown for disablility filter takes up the whole width of screen_


Cheese_Sat-05Oct24_11 26
_I'll work on this part next_

namanhboi (approved):
https://github.com/cornell-dti/carriage-web/pull/507#issuecomment-2395092578

PR #510, opened 2024-04-05
Added filter for disability in Student page

namanhboi:
### Summary 

Added filter for disability in Student page. Works with search bar and the show active/inactivate student option. As there are currently no designs for this filter box or any requirements on what to filter, I chose to filter by disability.

image
image

https://github.com/cornell-dti/carriage-web/assets/61220757/31628349-e3eb-43d9-b519-ffc757294fe2

### Test Plan 

I tested it manually on different inputs and it seems to work correctly. I don't know much about the correct (or any) approach to testing in this case so any input on this area would be greatly appreciated.

### Notes 

Everything seems to work fine.

### DIFF
--- a/frontend/src/components/UserTables/StudentsTable.tsx
+++ b/frontend/src/components/UserTables/StudentsTable.tsx
@@ -6,6 +6,8 @@ import { Row, Table } from '../TableComponents/TableComponents';
 import { useRiders } from '../../context/RidersContext';
 import styles from './table.module.css';
 import axios from '../../util/axios';
+import { Label } from '../FormElements/FormElements';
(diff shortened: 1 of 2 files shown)


Atikpui007 (changes requested):
Looks pretty good and I like that you are picking up Typescript and learning the industry conventions with typing. I however would like us to use the react-select library for all dropdowns to maintain consistency. 

So for now your PR will be blocked until the React-select library is merged

Atikpui007:
Merging this because I decided we dont need react-select capabilities for this selection dropdown as it is simply a filter

PR #512, opened 2024-04-14
Changed the success toast to reflect Lazim's design

namanhboi:
### Summary

Changed the design of the success toast to match this [design](https://www.figma.com/file/HBNXTvB6yYXXJqTgpjHl55/Lazim-Jarif---Carriage-Designs-(Spring-2024)?node-id=6%3A16896&mode=dev).

https://github.com/cornell-dti/carriage-web/assets/61220757/b3299665-7d78-456d-874b-8538739d5b8b

image

image

image

Note that the success modal will stay open even if you switch into a different page. I don't know if this is the intended behavior, but I can change it upon request. 

Also, in the file AuthManager.tsx, Toast was used directly, which I don't think is correct since we have toastContext which manages all the calling of the toasts. I changed the code to use toastContext and it compiled but I don't know how to see the changes made in the frontend. 

### Test Plan
 
I tested the success modal when adding new employees and locations, but I wasn't able to add students for some reason so I haven't been able to test it on that.  

I tried it on different screen sizes and on small screens:
image

The modal is too big, so I would need to do a media export in the future.


frontend/src/components/UserDetail/UserDetail.tsx
```diff
@@ -97,14 +97,14 @@ const UserDetail = ({
       });
     }
   };
+  if (isShowing && rider) {
+    showToast(
+      `Rider ${rider.active ? 'activated' : 'deactivated'}.`,
+      toastType ? ToastStatus.SUCCESS : ToastStatus.ERROR
```
Atikpui007, line 103: Since it has been changed to use the enum only lets refactor the toggleActive functionality to behave like this 

set the success and error in the then and catch clauses respectively 
So no need for an if statement 
showToast can be called directly in the then and catch clause.

frontend/src/components/ConfirmationToast/ConfirmationToast.tsx
```diff
@@ -1,19 +1,38 @@
...
 
-const Toast = ({ message, toastType }: toastProps) => {
+const Toast = ({ message, toastType, onClose }: toastProps) => {
   return typeof toastType === 'undefined' ||
     toastType == ToastStatus.SUCCESS ? (
     <div className={`${styles.toast} ${styles.successToast}`}>
-      <img alt="toast check" src={check} />
-      <p className={styles.toasttext}>{message}</p>
+      <button
+        onClick={() => {
```
Atikpui007, line 17: You can directly call onClose like this `onClick={onClose}`. Lets keep code consistency

frontend/src/components/ConfirmationToast/ConfirmationToast.tsx
```diff
@@ -1,19 +1,38 @@
...
+        className={styles.closeButton}
+      >
+        <img src={close} className={styles.closeImg}></img>
+      </button>
+      <div className={styles.toastContent}>
+        <img alt="toast check" src={green_check} />
+        <p className={styles.toasttext}>{message}</p>
+      </div>
+      <button
+        onClick={() => {
```
Atikpui007, line 29: Do same for above

Atikpui007 (changes requested):
Overall a pretty good PR. I left a couple of comments that should be addressed. Also the expected behavior of modals and by extension this success/ failure toast should be that when any other part of the page is clicked it goes away.

namanhboi:
I was able to implement what you suggested and now everthing outside the success modal is unclickable. However, I couldn't figure out a way to make the outside unscrollable.

Atikpui007 (dismissed):

namanhboi:
I merged it with the main branch and the success modal works as intended. 
Cheese_Wed-02Oct24_15 53
Cheese_Wed-02Oct24_15 54
Cheese_Wed-02Oct24_15 54

However, I discovered a few bugs with the master branch which I will point out in a separate pr

PR #516, opened 2024-05-06
Rides must be at least 5 minutes long

raissaji:
### Summary 

This pull request requires each ride to be at least 5 minutes long.

### Test Plan 

Screenshot 2024-05-05 at 6 13 03 PM

### Notes 

This PR does not require that each driver's rides are at least 5 minutes apart.


selenaliu1:
Hi Raissa! Overall this pull request looks good to me :) One thing I noticed is there is some repetition here of the following pattern: parseInt(time.substring(0, time.indexOf(':'))). This repeats here and for minDiff, so I think you could consider writing a helper function to abstract this for readability purposes.

PR #514, opened 2024-05-05
Filter out drivers in Reassign drivers for scheduled rides based on availability.

namanhboi:
### Summary 
When you reassign driver for scheduled rides, it will only show the drivers available. Additionally, you add can more filters into an array called reassignDriverFilters. 

image

### Test Plan

I have tested this very informally using console.log but plan to do more rigrous testing with unit tests in the future (after finals).

PR #517, opened 2024-05-06
No Overlapping Rides per Driver

raissaji:
### Summary 

This pull request is a WIP to ensure that each driver's scheduled rides are at least 5 minutes apart. Let's call the ride being created the impending ride. So far, the following has been implemented:

- The program identifies which drivers have rides that overlap with the impending ride's ride time
- Creates a console log saying which driver has a conflict

The following is the remainder of what needs to be implemented

- Remove the conflicting drivers from the DriverPage

### Test Plan 

Screenshot 2024-05-05 at 8 55 10 PM

Here, we scheduled one ride for driver bruh bruh at 9 AM - 9:20 AM and one ride for driver Ed von Aderkas at 1:50 PM - 3:11 PM.

Screenshot 2024-05-05 at 8 55 53 PM

Here, we add a ride from 9:04 AM - 9:15 AM. This impending ride should conflict with bruh bruh's schedule.

Screenshot 2024-05-05 at 8 56 18 PM

In the console log, bruh bruh indeed has a conflict. To complete this PR, we need to note the ID of every driver with a conflict and remove them from the available drivers displayed on the next page (the DriverPage).

### DIFF
--- a/frontend/src/components/RideModal/RideModal.tsx
+++ b/frontend/src/components/RideModal/RideModal.tsx
@@ -16,6 +16,11 @@ type RideModalProps = {
   editSingle?: boolean;
 };
 
+interface RideTime {
+  startTime: string;
+  endTime: string;
+}
+
 const RideModal = ({ open, close, ride, editSingle }: RideModalProps) => {
   const originalRideData = getRideData();
   const [formData, setFormData] = useState<ObjectType>(originalRideData);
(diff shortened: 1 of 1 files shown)


Atikpui007 (commented):
This is a good start towards the PR. Ideally we would not want to console.log but show errors on the page. 

An approach you should consider. 
- When querying the available drivers you can pass a ride prop into the availableDrivers and can filter drivers out based on the rides. The filtering has been abstracted so you can take a look at this pull request #514 to display drivers

PR #518, opened 2024-05-06
Added snapping when scrolling the schedule on homepage.

namanhboi:
### Summary 

Implemented snapping when scrolling on the schedule. The the rows are snapped on top and the columns are snapped to the end.

[screen-capture (2).webm](https://github.com/cornell-dti/carriage-web/assets/61220757/281684a6-625c-4e82-a0ea-20c2e4c91c5a)

The vertical scrolling works as expected, but there's still some weird behavior for horizontal scroll. The scroll sometimes get snapped to a portion in the middle of a cell instead of to the end. I will try to investigate the docs for scroll-snap-align more thoroughly in the future to see why.

[screen-capture (3).webm](https://github.com/cornell-dti/carriage-web/assets/61220757/67c425e5-1d7f-4749-b606-4963552feb7d)
_Some weird behavior with horizontal snapping_

This pull request is the first step towards redesigning the schedule on the home page.

### Test Plan 

I tested the scrolling on different screen sizes and it doesn't work too well on smaller screens like the Iphone, so I might have to do a media queries in the future.

### DIFF
--- a/frontend/src/components/Schedule/big_calendar_override.css
+++ b/frontend/src/components/Schedule/big_calendar_override.css
@@ -16,7 +16,7 @@
 
 /* other seperating timeslots */
 .rbc-day-slot .rbc-time-slot {
-  border-top: 0.063rem solid rgba(0, 0, 0, 15%); 
+  border-top: 0.063rem solid rgba(0, 0, 0, 15%);
 }
 
 /* line below drivers */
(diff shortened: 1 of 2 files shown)


Atikpui007:
closing because we can handle this in a better way. Will update the ticket to address this issue

PR #536, opened 2024-10-06
dka36/Reassign Button Task

Dwain-Anderson:
### 
Summary
ReAssign Button
- Currently, if Assign is clicked while the assign window is already opened, it won't go away on click. 
- Change it so that if assign is clicked while the assign window is already opened, it would close it. 

I fixed this issue by fixing the underlying race condition between function useOutsideAlerter(wrapperRef: any, buttonRef: any) in AssignDriverModal.tsx and const assignButton() in RidesTable.tsx

The function useOutsideAlerter called close() on any click outside the modal including the button, this would cause a change in useState() before the event handler for onclick in const assignButton() could properly read it's state (openAssignModal) to determine whether the modal was already opened. 

I fixed this by adding some logic to function useOutsideAlerter to check if the button was pressed as well, by passing a adding a new type to AssignModal reference, btn, as in in the button that triggered the modal window to open it. 

Tested on 1,2, 3, and 5 active ride entries in the table,
As well as tested on creations and deletions since I used mutability/button references.

### DIFF
--- a/frontend/src/components/Modal/AssignDriverModal.tsx
+++ b/frontend/src/components/Modal/AssignDriverModal.tsx
@@ -10,6 +10,7 @@ type AssignModalProps = {
   ride: Ride;
   allDrivers: Driver[];
   reassign: boolean;
+  buttonRef: any
 };
 
 type DriverRowProps = {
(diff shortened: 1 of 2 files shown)


Atikpui007 (approved):
Great catch. React does have a weird way in handling rendering when there are multiple events happening at the same time. Nice work

PR #537, opened 2024-10-09
Employee Card styling

selenaliu1:
### Summary 

This pull request fixes the styling on the employee card modal so that words aren't cut off and removes availabilities from the display. 

### Test Plan 
Before: 
Screenshot 2024-10-09 at 6 08 06 PM

After:
Screenshot 2024-10-09 at 6 06 48 PM

### Notes 

I commented out the availability information because I think it should be used somewhere else. Clicking on the employee card just redirects to the home page. Also uploading a picture for the profile picture doesn't work. 

### Breaking Changes 

### DIFF
--- a/frontend/src/components/Card/card.module.css
+++ b/frontend/src/components/Card/card.module.css
@@ -1,7 +1,7 @@
 .card {
   display: inline-block;
   width: 17.2rem;
-  height: 23rem;
+  /* height: 23rem; */
   background-color: white;
   border-radius: 0.5rem;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


raissaji (commented):
Hey Selena, your code is straightforward and gets the job done! I tested your code and there is no employee information that bleeds off the cards anymore, which is great. I agree that their availabilities should be displayed somewhere else, and I think it can go in a pop up window that expands the employee card when you click on it.

Atikpui007:
This is great work. I tested it out and everything seems to function as expected. 
Regarding the other bugs you brought up these will be addressed in other PRs

Atikpui007 (approved):

PR #539, opened 2024-10-16
WIP: Need to make recurring rides work as a feature; Fixed Schedule so that rides from other days show up.

namanhboi:
### Summary 
Currently, recurring rides doesn't work at all. It doesn't show up in the scheduled/unscheduled rides table of other days nor does it show up in the big Schedule in the home page.
[screen-capture (3).webm](https://github.com/user-attachments/assets/64c0f907-702c-4c69-b3c6-94460215635a)

This is because the useRides() function that both the rides table and schedule uses from RidesContext.tsx only pulls from the backend the rides from the current day. Since the recurring rides doesn't exist in the backend, they won't show up in the tables or the schedule. 

I plan to add the recurring rides feature by updating ```scheduledRides``` and ```unscheduledRides``` of ```RidesContext``` to have both the normal and recurring rides, which haven't been added to the backend yet. This ensures that these recurring rides show up in the ride tables and schedules. When we need to delete or edit these recurring rides, that's when we communiate with the database. 

Additionaly, I also found 2 bugs which I both fixed: 
- The schedule doesn't show rides from days other than the original day.
Cheese_Wed-16Oct24_03 09
__Here, the ride at the bottom doesn't show up on the schedule__
- Can't edit recurring rides.
Cheese_Wed-16Oct24_03 09_000
__I clicked on the edit button of the ride on the bottom, which is a recurring ride, and nothing shows up.

However, I found another bug in the backend that diasbled access to the carriage website which persists on the master branch.
Cheese_Wed-16Oct24_03 14_000
Cheese_Wed-16Oct24_03 14
The origin is here
Cheese_Wed-16Oct24_03 16
This might have been caused by me doing some work in ```RidesContext```. 

This pull request is the first step towards making the recurring rides work.
- [x] added recurring rides to ```scheduledRides``` and ```unscheduledRides``` of ```RidesContext``` . These rides are not in the back end.
- [ ] need to add a new ride to the backend when you edit a recurring ride 
- [ ] need to do something about deleting a recurring ride (delete all recurring rides or just that one).
- [x] fixed the 2 bugs mentioned above
 
### Test Plan 
I plan to add recurring rides and manually check if they show up in other days.

### Breaking Changes 
This bug: 
Cheese_Wed-16Oct24_03 14_000
This might have been caused by me doing some work in ```RidesContext```. The master branch now doesn't work?

### DIFF
--- a/frontend/src/context/RidesContext.tsx
+++ b/frontend/src/context/RidesContext.tsx
@@ -1,8 +1,9 @@
-import React, { useCallback, useState, useRef, useEffect } from 'react';
+import React, { useCallback, useState, useRef, useEffect, useContext } from 'react';
 import { Ride, Type } from '../types';
 import { useDate } from './date';
 import { format_date } from '../util/index';
 import axios from '../util/axios';
+
 
 type ridesState = {
   unscheduledRides: Ride[];
@@ -23,6 +24,15 @@ type RidesProviderProps = {
   children: React.ReactNode;
 };
 
+function uuidv4() {
+  return '10000000-1000-4000-8000-100000000000'.replace(/[018]/g, (c) =>
+    (
+      +c ^
+      (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (+c / 4)))
+    ).toString(16)
+  );
+}
+
 export const RidesProvider = ({ children }: RidesProviderProps) => {
   const [unscheduledRides, setUnscheduledRides] = useState<Ride[]>([]);
   const [scheduledRides, setScheduledRides] = useState<Ride[]>([]);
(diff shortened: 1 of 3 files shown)


frontend/src/context/RidesContext.tsx
```diff
@@ -23,23 +24,101 @@ type RidesProviderProps = {
...
 
   const refreshRides = useCallback(async () => {
-    const ridesData: Ride[] = await axios
+    const ridesDataToday: Ride[] = await axios
       .get(`/api/rides?date=${date}`)
       .then((res) => res.data)
       .then((data) => data.data);
-    if (ridesData) {
+
+    const newRecurringRides: Ride[] = await axios
```
raissaji, line 45: I believe the API is returning undefined data to ridesDataToday. An idea for error handling is to use a try catch to log what data you are fetching and see if it expected or not.

raissaji (commented):
Hi Nam, great job noticing the bugs! When I pulled your branch, I am coming across an error 500—I took a look at the server log and it appears that the problem lies in your get request `const ridesDataToday: Ride[] = await axios` in RidesContext. I left a more specific comment in the file that could help you with debugging.

Because of the AxioError, I couldn't add a ride to test how you squashed the bugs.

namanhboi:
Plan to implement recurring rides in the next commit: 

Recurring rides design: 

- source rides (the original recurring ride) is the current parentId field in Ride type.  
- all rides have the immediate parent id: immediateParentId? (need to add this field to Ride type)  
- children’s id (only one): childrenId (need to add this field to Ride type)

1. never return an original recurring ride from useRides()  
- all recurring rides returned from useRides() are alias rides: all rides object with recurring \= true has parentId that is not null (they all link to some ride in the db). 

2. 3 options to edit or delete:  
- Edit:  
  1. edit all :   
- Go to the source ride, delete all children ride by recursing down the children’s id.   
- apply edit to original   
- refreshRides  
  2. edit from now until the end date of current ride:  
- trim the end date of the immediate parent to before today  
- go down the tree of all children rides and delete all of them   
- apply edit to the ride.  
- refreshRides  
  3.  edit a single ride:   
- trim end date of immediate parent to before this day.  
- add a new ride on that date  
- create a new ride on the date \+ 1 with data similar to parent if the parent’s endTime allow for it. Add the id of the parent ride as the id of this ride’s parentId.  
- refreshRides  
- Delete:   
1. delete all:   
- go to source ride, delete all children rides  
- refreshRides  
2. delete from now until end date of current ride  
- go to parent ride, delete all children rides  
- refreshRides  
3. delete this single ride:   
- trim end date of immediate parent to before this day.  
- delete all children rides of the immediate parent.   
- create a new ride on the date \+ 1 with data similar to parent if the parent’s endTime allow for it. Add the id of the parent ride as the id of this ride’s parentId.  
- refreshRides

namanhboi:
Bug report: 
This ride seem to cause error in the master branch. IDK why
Cheese_Wed-30Oct24_15 35

namanhboi:
This ride causes error on the server side because the rider is a name instead of an id to a driver object in Dynamo.
Cheese_Wed-30Oct24_15 42

Gotta be mindful of this when creating recurring rides.

PR #544, opened 2024-10-23
dka36/HolidaysEnum

Dwain-Anderson:
Implemented Holidays Enum feature: 
image
Holidays now are blanked out on the calendar.


frontend/src/components/MiniCal/MiniCal.tsx
```diff
@@ -5,34 +5,20 @@ import 'react-datepicker/dist/react-datepicker.css';
...
-
 const isHoliday = (date: Date) => {
-  for (const holiday of holidays) {
-    if (holiday.startDate <= date && date < holiday.endDate) {
-      return true;
-    }
-  }
-  return false;
+  return holidaysList.some(
+    (holiday) => holiday.startDate <= date && date < holiday.endDate
```
selenaliu1, line 16: Hey Dwain! I checked out your branch and I like how Thanksgiving and Winter Break are greyed out on the calendar. However, I think the logic here might be off, because I can't see the other holidays. For one day holidays, try date <= holiday.endDate? Also, be careful about what dates you're actually comparing here, because it's possible it's comparing the dates down to the minute/second instead of the days themselves, so account for that as well.

Dwain-Anderson: Thanks for the catch Selena.

Atikpui007:
Pushed a Prettier fix so the CI Check should pass now

Atikpui007 (approved):
Includes rebased changes. 
Main change being the Parsing of Holidays 
Works as intended

PR #545, opened 2024-10-26
Implement Employee Card Pagination

selenaliu1:
### Summary 

Using Material UI, add Pagination component to the employee page so that instead of scrolling, users can click through the pages to find admins and drivers.

### Test Plan 
Before: (scrolling)

https://github.com/user-attachments/assets/edefb3e7-105d-46c6-9924-fb364531234b

After: (8 employees/page) 
 

https://github.com/user-attachments/assets/9a53dcea-ca60-4e12-afbc-fc772c63f53e

### Notes 

Had a breaking error --- was resolved by running npm install in carriage-web/frontend (didn't work just by running npm install in carriage-web).

### Breaking Changes 

### DIFF
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -8,6 +8,7 @@ import { useEmployees } from '../../context/EmployeesContext';
 import { AdminType } from '../../../../server/src/models/admin';
 import { DriverType } from '../../../../server/src/models/driver';
 import { Button } from '../FormElements/FormElements';
+import Pagination from '@mui/material/Pagination';
 
 const formatPhone = (phoneNumber: string) => {
   const areaCode = phoneNumber.substring(0, 3);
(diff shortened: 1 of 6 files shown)


Atikpui007:
This is weird because it works for me

I just pulled your branch and checked and everything was working fine


https://github.com/user-attachments/assets/9fbd5dfc-a130-4073-b22f-e564112452f7

It might be a case you havent installed MUI correctly but send me a copy of the error logs and lemme see

frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -74,21 +75,21 @@ const EmployeeCard = ({
    * @param availability the driver's availability, represented as an object map of days to start and end times
```
Atikpui007, line 75: You can delete this since we arent displaying it on the card anymore no need to comment it out

Atikpui007 (commented):
This is a good PR and adding of the MUI Pagination Component. 

One thing I would like to say though is that instead of displaying as single row per page you can extend it so that it is 2 rows per page or what would be most aesthetic. Additionally the Pagination should be more present at the bottom instead of right below the employee cards 


You can use a flex box for the container of the employee cards and grow it while the pagination will be fixed at the bottom

Atikpui007 (approved):
Glad we could get this fixed. This looks good and it works as expected. Tested it on couple browsers. 
Great job

PR #558, opened 2024-12-02
AWS S3 Bucket Photo Uploads and API Bug Fixing

Dwain-Anderson:
Implemented the ability to add employee photos using a newly created AWS-S3 bucket.
Revamped existing upload functions and added checks for existing API.

##### Description:
This endpoint allows users to upload images associated with Drivers or Admins and stores the images in an Amazon S3 bucket. It then updates the respective database record with the uploaded image's URL.

##### Request Headers:

| Header       | Type   | Description                  |
|--------------|--------|------------------------------|
| `Authorization` | String | Required. User authentication token for validating access. |

##### Request Body:
The request body should be sent in JSON format with the following fields:

| Field       | Type   | Description                                                      |
|-------------|--------|------------------------------------------------------------------|
| `id`        | String | **Required.** Unique identifier for the Admin or Driver.         |
| `tableName` | String | **Required.** Specifies the target table: either `Drivers` or `Admins`. |
| `fileBuffer`| String | **Required.** Base64-encoded string representing the image file to be uploaded. |

###### **Success (200):**
If the image is successfully uploaded and the database is updated:
```json
{
  "message": "Image uploaded and database updated successfully.",
  "photoLink": "https://carriage-images.s3.us-east-2.amazonaws.com/<Drivers|Admins>/<id>"
}
```

###### Error Responses:

| Status Code | Description                              | Example Response                                                                 |
|-------------|------------------------------------------|---------------------------------------------------------------------------------|
| 400         | Invalid or missing `id`                 | `{ "err": "Invalid ID: empty or missing" }`                                     |
| 400         | Invalid or missing `fileBuffer`         | `{ "err": "Invalid file buffer: empty or missing" }`                            |
| 400         | Invalid `tableName`                     | `{ "err": "Invalid table name: <tableName>" }`                                  |
| 400         | S3 upload error                         | `{ "err": "Error message from S3" }`                                           |
| 500         | General server error                    | `{ "err": "Unexpected server error occurred" }`                                |

##### Validation Rules:
1. **ID Validation:**
   - Must be a non-empty string.
2. **File Buffer Validation:**
   - Must be a non-empty Base64-encoded string.
3. **Table Name Validation:**
   - Must be either `Drivers` or `Admins`.

---

##### Implementation Details:
- **AWS S3 Bucket:** Images are uploaded to the S3 bucket named `carriage-images` in the `us-east-2` region.
- **Object Key:** Images are stored under `<tableName>/<id>`.
- **Access Control:** Uploaded images are publicly readable (`public-read`).
- **File Size Limit:** Images can be up to **50,000 MB** in size.
- **Content Encoding:** The `fileBuffer` is decoded from Base64 before being sent to S3.

Usage Notes
- **Database Update:** After the image is successfully uploaded, the corresponding `photoLink` is saved in the database for the given Admin or Driver.
- **Permissions:** Only authenticated users with the `User` role can access this endpoint. The uploaded image is only allowed for `Drivers` and `Admins`.

WIP/next iteration
Weird availability parse error (unrelated to this api I think)
Better cache (unsure if needed)

### DIFF
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -5,14 +5,14 @@ import { phone, wheel, user } from '../../icons/userInfo/index';
 import { Employee } from '../../types';
 import { AdminType, DriverType } from '../../types';
 
-const formatPhone = (phoneNumber: string |undefined) => {
+const formatPhone = (phoneNumber: string | undefined) => {
   if (phoneNumber !== undefined) {
     const areaCode = phoneNumber.substring(0, 3);
     const firstPart = phoneNumber.substring(3, 6);
     const secondPart = phoneNumber.substring(6, 10);
     return `${areaCode}-${firstPart}-${secondPart}`;
   } else {
-    console.error("Undefined PhoneNumber")
+    console.error('Undefined PhoneNumber');
     return '';
   }
 };
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -5,11 +5,16 @@ import { phone, wheel, user } from '../../icons/userInfo/index';
 import { Employee } from '../../types';
 import { AdminType, DriverType } from '../../types';
 
-const formatPhone = (phoneNumber: string) => {
-  const areaCode = phoneNumber.substring(0, 3);
-  const firstPart = phoneNumber.substring(3, 6);
-  const secondPart = phoneNumber.substring(6, 10);
-  return `${areaCode}-${firstPart}-${secondPart}`;
+const formatPhone = (phoneNumber: string |undefined) => {
+  if (phoneNumber !== undefined) {
+    const areaCode = phoneNumber.substring(0, 3);
+    const firstPart = phoneNumber.substring(3, 6);
+    const secondPart = phoneNumber.substring(6, 10);
+    return `${areaCode}-${firstPart}-${secondPart}`;
+  } else {
+    console.error("Undefined PhoneNumber")
+    return '';
+  }
 };
 
 type EmployeeCardProps = {
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -86,7 +86,7 @@ const EmployeeCard = ({
         firstName={firstName}
         lastName={lastName}
         netId={netId}
-        photoLink={photoLink}
+        photoLink={photoLink ? photoLink : 'frontend/public/logo512.png'}
       >
         <CardInfo icon={phone} alt="phone">
           <p>{fmtPhone}</p>
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -74,7 +74,7 @@ const EmployeeCard = ({
         firstName={firstName}
         lastName={lastName}
         netId={netId}
-        photoLink={photoLink}
+        photoLink={photoLink ? `${photoLink}?t=${new Date().getTime()}` : ''}
       >
         <CardInfo icon={phone} alt="phone">
           <p>{fmtPhone}</p>
--- a/frontend/src/components/EmployeeCards/EmployeeCards.tsx
+++ b/frontend/src/components/EmployeeCards/EmployeeCards.tsx
@@ -86,7 +86,8 @@ const EmployeeCard = ({
         firstName={firstName}
         lastName={lastName}
         netId={netId}
-        photoLink={photoLink ? photoLink : 'frontend/public/logo512.png'}
+        photoLink={
+          photoLink }
       >
         <CardInfo icon={phone} alt="phone">
           <p>{fmtPhone}</p>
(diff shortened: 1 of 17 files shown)


frontend/src/components/EmployeeCards/EmployeeCards.tsx
```diff
@@ -33,7 +33,12 @@ const EmployeeCard = ({
 }: EmployeeCardProps) => {
   const navigate = useNavigate();
   const netId = email.split('@')[0];
-  const fmtPhone = formatPhone(phoneNumber);
+  const fmtPhone = '';
+  if (phoneNumber !== undefined) {
+    const fmtPhone = formatPhone(phoneNumber);
+  } else {
+    const fmtPhone = '';
```
selenaliu1, line 40: I dont think this else statement is needed since you set fmtPhone to '' already on line 36


also, nit: try to keep all the logic for formatPhone within the helper function, ie. do the check for if the phone number is undefined within formatPhone itself, so that if the helper is used in the future, you don't have to repeat this logic

frontend/src/components/EmployeeModal/Upload.tsx
```diff
@@ -1,34 +1,39 @@
...
-    if (files && files[0] && files[0].size < 500000) {
-      const file = files[0];
-      const photoURL = URL.createObjectURL(file);
-      setImageURL(photoURL);
+    if (files && files[0] && files[0].size < IMAGE_SIZE_LIMIT) {
+      setImageURL(URL.createObjectURL(files[0]));
+      imageChange(e);
     } else {
-      console.log('invalid file');
+      setErrorMessage(`Images must be under ${IMAGE_SIZE_LIMIT / 1000} KB`);
```
selenaliu1, line 34: Hey Dwain, I checked out this branch and was having some issues uploading an image at first, so I checked the logs and got "Images must be under 50 KB," which seems to have come from here. After choosing a smaller image (20 KB) I was able to upload an image but the first one I tried was 116 KB, which is a size I think users should also be able to use to upload, so either here or in a future PR, maybe try either increasing the `IMAGE_SIZE_LIMIT` or investigating why the if statement would be returning false here.
Otherwise, good work on figuring out the backend and getting the uploading images to work!!

frontend/src/components/EmployeeCards/EmployeeCards.tsx
Dwain-Anderson: Thanks, I'll fix it here. I left it in when debugging some error regarding workingContext sending duplicated request.

frontend/src/components/EmployeeModal/Upload.tsx
Dwain-Anderson: Hello, the img size limit was set prior, in server/app.ts its set to 5000kb and anything above it throws a payload size error. I can ask Desmond about changing that directly. Otherwise I'll look into other ways to get past it when i add the optimizations.

Dwain-Anderson:
-Fixed a memory leak with WorkingContext/Week Context where is would constantly overwrite state in the EmployeeModal API upload process.
-Increased image upload limit:
image

-More debugging, maybe re-looking at the db schema or enforcing some consistency for input/out as a good chunk of the current db entries differ in fields.

selenaliu1:
Hi Dwain,

Great work on this! I tested this on editing an employee profile with an image of size 106 KB, 4 MB, and 5.6 MB and they all worked. I also tried replacing the profile pictures with another picture, and it works, although you need to reload the page to see the updates. Really nice work!!

server/src/router/admin.ts
```diff
@@ -23,8 +24,9 @@ router.get('/', validateUser('Admin'), (req, res) => {
 // Put a driver in Admins table
 router.post('/', validateUser('Admin'), (req, res) => {
   const { body } = req;
+  const id = body.id || uuid();
```
selenaliu1, line 27: in what case is id being set to body.id?

selenaliu1 (commented):
Hi Dwain,

Great work on this! I tested this on editing an employee profile with an image of size 106 KB, 4 MB, and 5.6 MB and they all worked. I also tried replacing the profile pictures with another picture, and it works, although you need to reload the page to see the updates. Really nice work!!

server/src/router/admin.ts
Dwain-Anderson: I have this when a new admin is made, if its an existing driver it would keep their id from the drivers table, but if it's an entirely new employee then it generates a new id. Maybe drivers and admins shouldnt have the same ID if they have the same ID they would be the same person ig.

Dwain-Anderson:
- Fixed WorkingHours 
- Fixed issues with API uploading
- Fixed the server cacheing recent uploaded images, therefore not showing the most recent version immediately.
- Fixed sidebar profile not showing driver/admin profile pictures.
Quick Demo.
- Changed data processing in employee modals to use a more normalized version to reduce duplicate data, ideally this change would be implemented in the database and not just employee modal but it can wait.
https://github.com/user-attachments/assets/612cd96c-f7f3-4b06-814c-4d0dba4da298

selenaliu1 (dismissed):
Hey Dwain,

Really great work on this. I tested this on admins, drivers, and admin and drivers, and it worked as expected. Just a note that this doesn't seem to work on updating existing employee's photos, but that can come in a later PR/task. Nice job!

frontend/src/components/EmployeeCards/EmployeeCards.tsx
selenaliu1: Can you add back the check into the formatPhone function itself to make sure that the phone number is at least 10 digits / not undefined? this is very minor but there was a bug previously where phoneNumber was undefined in the backend and it was caused because formatPhone was called on an undefined phoneNumber

PR #574, opened 2025-04-13
Dka36/DriversModel

Dwain-Anderson:
### Summary

Updated the `Drivers` model in preparation for the new frontend page.

**Database Schema Change – `Drivers` Table:**  
Modified from:  
`Drivers(email, firstName, lastName, phoneNumber, photoLink, startDate, availability, vehicle)`  
To:  
`Drivers(email, firstName, lastName, phoneNumber, photoLink)`


server/src/router/driver.ts
```diff
@@ -67,54 +31,13 @@ router.get('/:id/profile', validateUser('User'), (req, res) => {
...
       firstName,
       lastName,
       phoneNumber,
-      availability,
-      vehicle,
-    });
-  });
-});
-
-// Get a driver's weekly stats
```
selenaliu1, line 83: why are we deleting this?

selenaliu1 (commented):
Looks good to me overall but I'm just curious, why are we getting rid of availability? Feels like that'd be important information to have but please lmk what I'm missing

server/src/router/driver.ts
Atikpui007: Conversations with Marcia ( The CULift ) dispatcher showed that Drivers are available all the time during the work day since it is considered a 9-5 job

Atikpui007 (approved):
Mainly removal changes for the Driver so looks good 
Doesnt currently cause any crashes

PR #585, opened 2025-05-04
Driver Card Availability

anika-4444:
### Summary 

- [Boom fix] changed driver card availability form -- only days of week asked for now

### Breaking Changes 


frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -13,13 +13,21 @@ import { useEmployees } from '../../context/EmployeesContext';
 import { useToast, ToastStatus } from '../../context/toastContext';
 import axios from '../../util/axios';
 
+enum DayOfWeek {
```
selenaliu1, line 16: do we need days of the week? my understanding is that all drivers work all weekdays

frontend/src/components/EmployeeModal/EmployeeModal.tsx
```diff
@@ -13,13 +13,21 @@ import { useEmployees } from '../../context/EmployeesContext';
...
+
 type AdminData = {
   type: string[];
   isDriver: boolean;
 };
 
 type DriverData = {
-  availability: any;
+  availability: DayOfWeek[];
   startDate: string;
```
selenaliu1, line 31: `export type DriverType = {
  id: string;
  firstName: string;
  lastName: string;
  phoneNumber: string;
  email: string;
  photoLink?: string;
};`

looks like this is the current DriverType, if we want to keep it like this then I think we can get rid of availability and startDate from the form altogether since it's not saved in the backend

selenaliu1 (approved):

PR #591, opened 2025-09-19
Create onboarding page for unregistered users

benkoppe:
### Summary 

If attempting to login as an unregistered user, display "you are not registered for Carriage" page.

- [x] create shared `UnregisteredUserType` for backend
- [x] define shared type as `UnregisteredUser` on frontend
- [x] implement `UnregisteredUserPage` frontend page
- [x] receive 'User not found' error on frontend and display page if necessary
- [x] extract login page `wavePath` to shared `WAVE_PATH` to help in replicating login UI

Screenshot 2025-09-19 at 3 03 07 AM

### Notes

This seems to be almost everything we need, but still some questions to consider before merging:

- [ ] Should more information be shared between the frontend/backend, such as the 'User not found' error message text that triggers this page? If not, this flow could be easy to break by accident.
- [ ] Should this onboarding page only display for "Student" login? Currently, for ease of implementation, it displays for all login types. If not, should other login types display nothing like before, or should we show different content on the same page structure?
- [ ] Should more frontend UI be shared between the login page and this new page? Currently the main piece of shared UI is the `WAVE_PATH`, but if we want to maintain the similarity between both pages (and add more in the future) we might want some type of shared layout between the pages.
  - This change would have the tradeoff that it would make it harder to modify just one page in the future.

### DIFF
--- a/frontend/src/components/Onboarding/UnregisteredUserPage.tsx
+++ b/frontend/src/components/Onboarding/UnregisteredUserPage.tsx
@@ -30,8 +30,8 @@ const UnregisteredUserPage: React.FC<UnregisteredUserProps> = ({
           </p>
           <p className={styles.contact}>
             Call <a href="tel:607-254-4545">607-254-4545</a> or email{' '}
-            <a href="mailto:carriage@cornell.edu">carriage@cornell.edu</a> if
-            you have any questions.
+            <a href="mailto:culift@cornell.edu">culift@cornell.edu</a> if you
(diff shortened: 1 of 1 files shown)


selenaliu1 (approved):
I was able to see the unregisteredUserPage when I removed myself as a driver and it was responsive, so nice job! I think it's fine to leave the message as generic User for now. I updated this PR to incorporate the SSO changes and made a couple changes so that it was compatible.

PR #594, opened 2025-09-22
WIP Favorites card

anika-4444:
### Summary 
This is a WIP for favoriting and un-favoriting rides. _Don't merge._ The CSS is implemented as well as a few backend axios changes. Currently, however, a 500 internal server error appears when the user attempts to add a favorite. I believe there might be a mismatch between the provided ride/user keys or the database schema

This pull request is the first step towards implementing a favorites section

- [x] implemented favorites button

### Test Plan 
favorites

### Notes 

It would be nice to implement the ability to edit a favorited ride, as right now you
can only favorite past rides (if there is one part of the past ride you'd like to change you'd have
to create a new ride then favorite it)


mjaydenkim:
Can we modify this PR to merge into Desmond’s branch instead of master? It might help show the actual changes of the PR instead of bundling all of Desmond’s changes with it.

frontend/src/components/RideDetails/RideOverview.tsx
```diff
@@ -205,6 +208,7 @@ const RideOverview: React.FC<RideOverviewProps> = ({ userRole }) => {
   const ride = editedRide!;
   const temporalType = getTemporalType(ride);
   const showRecurrence = userRole !== 'driver'; // Hide recurrence for drivers
+  const [isClicked, setIsClicked] = useState(false);
```
benkoppe, line 211: You might want a more descriptive variable name here than `isClicked`, something more descriptive about what the state represents.

frontend/src/components/RideDetails/RideOverview.tsx
```diff
@@ -313,6 +317,51 @@ const RideOverview: React.FC<RideOverviewProps> = ({ userRole }) => {
...
+   *
+   * TODO:
+   *  - Fix backend schema mismatch or ensure correct key format
+   *  - Add responsivity for button icon
+   */
+
+  const handleFavorite = async () => {
+    if (!isClicked) {
+      //user favorited icon
+      setIsClicked(true);
```
benkoppe, line 340: This function might have a mismatched state between `isClicked` and the actual result of the favoriting operation because of the catch statements below.

benkoppe (changes requested):
I like that you avoid touching the backend schemas too much, since this feature might definitely be blocked behind #598. Until then, things look pretty good, and introduction of a new state makes sense. It might need renaming, though, and we might want to double-check to make sure it always matches the server side.

PR #600, opened 2025-10-08
Implement email service

selenaliu1:
### Summary 

This pull request is the first step towards implementing an email service to automate emails when a ride is approved, approved with modification, rejected, or cancelled.
- [x] implemented nodemailer
- [x] added new scheduling state for approved rides with modifications
- [x] implement rejected vs cancelled state
- [x] misc clean up: removed status from student page and report button, make rides more clickable from admin page
- [x] made the cancel ride confirmation dialog uniform across all places needed on admin and student page, added reject ride option for admin
- [x] approved email
- [x] approved with modification logic and email flow
- [x] rejected email
- [x] cancelled email

### Test Plan 

Fermin created a VPN to test this on, I successfully received the emails! You won't get the emails unless you have access to the VPN and are connected to it.
Screenshot 2025-11-29 at 3 50 00 PM

Cancelled ride:
Screenshot 2025-11-20 at 10 35 42 PM

 rejected ride: 
Screenshot 2025-11-20 at 10 35 00 PM

scheduled with modification:
Screenshot 2025-11-20 at 10 36 03 PM

scheduled ride:
Screenshot 2025-11-20 at 10 36 58 PM

### Notes 

Kept console.logs in for now for debugging.

### Breaking Changes 

### DIFF
--- a/frontend/src/components/EmployeeModal/EmployeeModal.tsx
+++ b/frontend/src/components/EmployeeModal/EmployeeModal.tsx
@@ -12,6 +12,7 @@ import styles from './employeemodal.module.css';
 import { useEmployees } from '../../context/EmployeesContext';
 import { useToast, ToastStatus } from '../../context/toastContext';
 import axios from '../../util/axios';
+import { extractNetIdFromEmail } from 'util/userUtils';
 
 type AdminData = {
   type: string[];
@@ -109,7 +110,7 @@ const EmployeeModal = ({
       methods.reset({
         firstName: existingEmployee.firstName,
         lastName: existingEmployee.lastName,
-        netid: existingEmployee.netId,
+        netid: extractNetIdFromEmail(existingEmployee.email) || '',
         phoneNumber: existingEmployee.phoneNumber,
         startDate: existingEmployee.driver?.startDate,
         availability: existingEmployee.driver?.availability || [],
--- a/frontend/src/components/EmployeeModal/RoleSelector.tsx
+++ b/frontend/src/components/EmployeeModal/RoleSelector.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, Dispatch, SetStateAction, useState } from 'react';
+import React, { ChangeEvent, Dispatch, SetStateAction } from 'react';
 import styles from './employeemodal.module.css';
 import { Label, Input } from '../FormElements/FormElements';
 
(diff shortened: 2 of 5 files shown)


server/src/mailer.ts
```diff
@@ -0,0 +1,208 @@
...
+  newDropoff: string;
+  newTime: Date;
+}
+
+export async function sendApprovedEmail(to: string, ride: RideDetails) {
+  console.log('📧 MAILER: sendApprovedEmail called');
+  console.log('📧 MAILER: To:', to);
+  console.log('📧 MAILER: Ride:', ride);
+
+  const day = format(ride.time, 'EEEE');
```
mjaydenkim: Maybe add dates here? No full date is kinda ambiguous. e.g. "Your ride on Monday, _December 8,_ from 602 Tower Rd to 300 W College Ave at 16:04 was approved."

frontend/src/components/Modal/CancelRideConfirmationDialog.tsx
mjaydenkim: Looks great to me -- maybe we want to add driver (if assigned) and passenger to the ride summary?

server/src/router/ride.ts
```diff
@@ -447,13 +454,53 @@ router.put('/:id', validateUser('User'), (req, res) => {
...
+        // Send emails if needed (for scheduled, rejected, cancelled, or modified)
+        sendRideEmails(updatedRide, originalRide).catch((error) => {
+          console.error('Failed to send ride emails:', error);
+        });
+
         // send ride even if notification failed since it was actually updated
-        notify(ride, body, userType)
-          .then(() => res.send(ride))
-          .catch(() => res.send(ride));
+        notify(updatedRide, body, userType)
```
mjaydenkim: Updated rides seem to be sent as approved rides -- since these emails don't contain all the details this is a little confusing. Maybe add a new type of email for updated/modified rides? Probably not necessary though, might be more work than it's worth

mjaydenkim (commented):
This looks incredible, great work! Not approving just yet b/c I noticed a couple of small bugs including one seemingly breaking one, but this looks just about there -- great work <3 

Also, side note -- do we wanna add emails for when rides are assigned to drivers? I think it's less necessary but if it's easy to implement it might be worth it. Not sure though.

frontend/src/types/index.ts
```diff
@@ -88,6 +88,8 @@ export enum Status {
 // Scheduling state - separate from operational status
 export enum SchedulingState {
   SCHEDULED = 'scheduled',
+  SCHEDULED_WITH_MODIFICATION = 'scheduled_with_modification',
```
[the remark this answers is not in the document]
selenaliu1: Thanks for the catch! Kind of debating what use there is to have this scheduling state here in the first place, since it should be done automatically and could potentially cause confusion, but I'm hesitant to fully remove it in cases when automatic detection might not correctly catch a scheduling state change, or other edge cases I haven't fully thought through. The error should be fixed now, and the logic now is that even if the scheduling state changes to scheduled with modification, it doesn't send an email unless an actual change relevant to the rider (time, pickup or dropoff location) is made.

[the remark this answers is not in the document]
mjaydenkim: Yeah, I think that makes sense -- I think we should lean towards giving admins more leverage given how few there are and how flexible the current system is, but that solution sounds good to me.

server/src/router/ride.ts
mjaydenkim: That makes sense! I also noticed that changing the state back from "Scheduled_with_modifications" to "Scheduled" triggers a new email; I honestly don't think we need to worry about this though, it's a pretty small error.

mjaydenkim (approved):
Did some more testing and everything looks great! This is such incredible work, tysm :') left like one or two comments but there's no real concerns

PR #605, opened 2025-10-27
Replace Create React App with Vite

benkoppe:
### Summary 

This PR replaces CRA (Create React App) with Vite. Builds are significantly faster, especially in development mode, and Vite is generally more extensible and convenient. It also appears to be very useful for PR #604 due to its native typescript support (more testing needed). 

This change was very easy to do and appears to have only upsides, so it might be a good idea to go forward with. However, it is also totally unnecessary in general (barring what's mentioned above about its native ts support), so any hesitation is understood. 

But it would be nice, because seriously builds are so much faster. For example, wait time on an initial run of `npm run start`/`npm run dev` on a new branch went from ~30 seconds to instant.

- [x] add vite dependencies
- [x] change files over to vite standards
- [x] remove create react app dependenices
- [ ] rigorous testing to ensure everything still works

### Test Plan 

Tests with this change will need to be mostly manual, because Vite intends to work as a mostly transparent replacement for the same things once done by create react app.

### Breaking Changes 

- IMPORTANT: In dev mode, we must now type `npm run dev` rather than `npm run start`.
- IMPORTANT: All env variables in `.env` must be named to replace prefix `REACT_APP` with `VITE`
- I removed some of the frontend scripts, `test` and `eject`, as I believe eject is only a create react app thing, and I didn't add Vitest.


benkoppe:
I don't know why the changes to the github workflow are in here. That might need to be addressed.

PR #609, opened 2025-11-02
Fix time validation, standardize request modals

mjaydenkim:
### Summary

Fixed next/current ride edit button and ride table on-click function to lead to the same ride detail/edit modal; added consistent time validation when editing rides. 

image

### Test Plan
Tested through rigorous edge case input checking; anyone reviewing this PR should definitely do some of that on their own too to verify correctness.

### DIFF
--- a/frontend/src/components/AddRideButton/AddRideButton.tsx
+++ b/frontend/src/components/AddRideButton/AddRideButton.tsx
@@ -1,9 +1,9 @@
 import React, { useState } from 'react';
-import { Button } from '../FormElements/FormElements';
 import RideDetailsComponent from '../RideDetails/RideDetailsComponent';
 import { createEmptyRide } from '../../util/modelFixtures';
 import { RideType } from '../../types';
 import { useRides } from '../../context/RidesContext';
+import buttonStyles from '../../styles/button.module.css';
 
 const AddRideButton: React.FC = () => {
   const [open, setOpen] = useState(false);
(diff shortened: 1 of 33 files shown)


frontend/src/components/RideDetails/RideActions.tsx
```diff
@@ -92,6 +93,8 @@ const RideActions: React.FC<RideActionsProps> = ({
   const [selectedStatus, setSelectedStatus] = useState<Status | null>(null);
   const [updating, setUpdating] = useState(false);
   const [saving, setSaving] = useState(false);
+  const [timeErrorOpen, setTimeErrorOpen] = useState(false);
+  const [timeErrors, setTimeErrors] = useState<TimeValidationError[]>([]);
```
benkoppe: Do `timeErrorOpen` and `timeErrors` need to be independent pieces of state? Can one be derived with a `useMemo`?

frontend/src/components/RideDetails/RideActions.tsx
```diff
@@ -159,6 +162,24 @@ const RideActions: React.FC<RideActionsProps> = ({
...
   const handleSave = async () => {
+    // Validate times before saving; show modal with errors if invalid
+    const allowPastTimes = false;
+    const timeValidation = validateRideTimes(
+      ride.startTime,
+      ride.endTime,
+      {
+        allowPastTimes,
+        maxDurationHours: 24,
+        minDurationMinutes: 5,
```
benkoppe: How was 5 minutes chosen as the minimum? We might want to just make this 0 or 1 so we never get in the way on minimums.

frontend/src/components/RideDetails/RideActions.tsx
```diff
@@ -527,6 +548,34 @@ const RideActions: React.FC<RideActionsProps> = ({
         </DialogActions>
       </Dialog>
 
+      {/* Time Validation Errors Modal */}
```
benkoppe: I might have missed when this was already discussed, but what was the reasoning behind a modal specifically for these validation errors? Wouldn't something inline on the layout be a little more clear, to avoid stacking modals on top of modals, or are we switching due to layout shift and consistency issues? Either way it might be useful if the PR summary had justification behind this consideration.

benkoppe (commented):
I'm a little iffy about this change, because as someone with a certified trash memory ™ I'm the kind of person that would need to press the submit button multiple times to remember what I did wrong because the error text doesn't persist 😭. Still an understandable change for various benefits, but maybe some discussion about the tradeoffs is needed

PR #618, opened 2025-11-22
Locations textfield

anika-4444:
### Summary 

- [x] implemented custom location textfield: users can now select 'other' and a Google Maps search bar pops up. An address can then be entered and a pin on the map appears. 
- [x] added documentation to ride request dialog
- [x] added api route enabling users to create locations 
- [x] fixed small bug regarding start and end location. Previously, if a user selected the end location first, they could pick the same location to be dropped off at. By disabling drop-off until pick-up is selected, this circumvents the problem 
- [x] changed LocationsContext to ensure custom locations do not appear within Carriage
- [x] custom location only added if it does not already exist within the Locations table in Carriage database

### Breaking Changes 

### DIFF
--- a/frontend/src/components/RiderComponents/RequestRideDialog.tsx
+++ b/frontend/src/components/RiderComponents/RequestRideDialog.tsx
@@ -791,7 +791,7 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
                     )}
                     {!formData.pickupLocation && (
                       <FormHelperText>
-                        {'Please select pickup location'}
+                        {'Please choose pickup location first'}
                       </FormHelperText>
                     )}
                   </FormControl>


frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -91,7 +106,9 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
   supportedLocations,
   ride,
 }) => {
-  const { locations } = useLocations();
+  //official locations with other added
+  const supportLocsWithOther = [...supportedLocations, Other];
```
YottaYocta: I feel this is pretty minor, but do you think that moving Other to the front of the list would help with UX? If `supportedLocations` get long enough I'm worried that users will have to scroll to specify a custom location.

Then again, 'other' as a first option would be kinda odd from a wording perspective. If you think having it be the first option is a good idea, maybe call it something like 'custom'?

frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -196,6 +232,160 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
...
+   * 1. Normalizing the address
+   *
+   * 2. Searching for an existing Location via address or lat/lng matching
+   *
+   * 3. Creating a new Location if needed
+   * This prevents multiple users from creating duplicate “custom” entries
+   * for the exact same geolocation.
+   *
+   */
+  const createOrGetLocation = async (
```
YottaYocta: image
Eeyyy nice 👍

server/src/router/location.ts
YottaYocta: 👍

frontend/src/context/LocationsContext.tsx
YottaYocta: 👍

YottaYocta (approved):

PR #626, opened 2025-12-05
Rwq3/small frontend fixes

YottaYocta:
### Summary 

This PR makes many small changes in styling and functionality. It fixes a couple bugs with admin timeline view

- [x] extracts button stylings used in student side to general css module that's usable across project
- [x] adds functionality to export scheduled, unscheduled, or combined rides for a given day

image

produces...

image

- [x] improves the styling of buttons currently
- [x] removes notification bell across admin pages
- [x] replaces all primary and second buttons on top row (except for analytics) with consistently styled buttons
- [x] zoom in/out on timeline
- [x] styling cleanup, removing unused files and vars
- [x] new day-based navigation on admin dash   
- [x] many other minor styling tweaks 

# Before
image

# After
image

# Before
image

# After
image

# Before
image

# After
image

### DIFF
--- a/frontend/src/components/DayNavigation/DayNavigation.tsx
+++ b/frontend/src/components/DayNavigation/DayNavigation.tsx
@@ -11,20 +11,12 @@ export const DayNavigation = () => {
     const newDate = new Date(curDate);
     newDate.setDate(newDate.getDate() - 1);
 
-    while (newDate.getDay() === 0 || newDate.getDay() === 6) {
-      newDate.setDate(newDate.getDate() - 1);
-    }
-
     setCurDate(newDate);
   };
 
   const nextBusinessDay = () => {
     const newDate = new Date(curDate);
     newDate.setDate(newDate.getDate() + 1);
-
-    while (newDate.getDay() === 0 || newDate.getDay() === 6) {
-      newDate.setDate(newDate.getDate() + 1);
-    }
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -116,11 +119,14 @@ const ScheduledTimeline: FC<ScheduledTimelineProps> = ({
...
-  const minuteWidth = useMemo(() => halfHourWidth / 30, [halfHourWidth]);
+  const minuteWidth = useMemo(
+    () => localHalfHourWidth / 30,
+    [localHalfHourWidth]
+  );
 
   const baseTime: Date = useMemo<Date>(() => {
     const baseTimeForToday = new Date(baseDate);
-    baseTimeForToday.setHours(7);
+    baseTimeForToday.setHours(0);
```
mjaydenkim: I like that the entire day fits inside the window now, but that also means that the first view starts at midnight to around 5am when rides will typically never be scheduled -- wonder if we can set the default to partway through the schedule?

mjaydenkim (approved):
I have no real notes -- this looks great! I wasn't able to test is as comprehensively as I want to b/c I'm tired but I don't see any major red flags. I think it's good to go! New styles look incredible + new detailed exports

PR #629, opened 2025-12-07
WIP: Custom Location Enhancements

anika-4444:
### Summary 

- [x] implemented better date and time validation
- [x] the ability to overlay CU-Lift's operation radius, updates dynamically with CU-Lift updates to Google Maps file
- [x] better location checking

### Test Plan 

image

### Notes 

The KML link is not CU-Lift's actual radius right now because I did not have the proper permissions to export the file. Also, I plan on adding location validation in this PR that checks whether or not a custom location is actually allowed 

### Breaking Changes 

### DIFF
--- a/frontend/src/components/RiderComponents/RequestRideDialog.tsx
+++ b/frontend/src/components/RiderComponents/RequestRideDialog.tsx
@@ -964,29 +964,28 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
                   )}
                 </div>
               </div>
-
-              <div className={styles.mapColumn}>
-                <div style={{ marginBottom: '12px' }}>
-                  <FormControlLabel
-                    control={
-                      <Switch
-                        checked={overlayRadius}
-                        onChange={(e) => setOverlayRadius(e.target.checked)}
-                      />
-                    }
-                    label="CULift Service Radius"
-                  />
-                </div>
-
-                <RequestRideMap
-                  pickupLocation={safePickup}
-                  dropoffLocation={safeDropoff}
-                  availableLocations={getAvailableLocations()}
-                  onPickupSelect={handleLocationSelect}
-                  onDropoffSelect={handleLocationSelect}
-                  kmlLink={kmlLink}
+            </div>
+            <div className={styles.mapColumn}>
+              <div style={{ marginBottom: '12px' }}>
+                <FormControlLabel
+                  control={
+                    <Switch
+                      checked={overlayRadius}
+                      onChange={(e) => setOverlayRadius(e.target.checked)}
+                    />
+                  }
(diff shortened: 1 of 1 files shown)


frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -431,12 +452,31 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
   const handleDateChange =
     (field: keyof Pick<FormData, 'date' | 'time' | 'repeatEndDate'>) =>
     (newDate: Date | null) => {
+      if (field === 'time') {
```
benkoppe: Would it be better to separate time changes into a separate function from date changes?

frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -431,12 +452,31 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
...
+        const isValid =
+          newDate !== null
+            ? isValidTime(newDate)
+            : setFormData({
```
benkoppe: I'm confused by the flow of this function. If working with a time field, does this mean you call `setFormData` from within the ternary statement, and then also call `setFormData` again lower down? I think there may be a better way of doing things.

frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -431,12 +452,31 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
...
+        setIsInvalidTime(!isValid);
+      }
+
       setFormData({
         ...formData,
         [field]: newDate,
       });
     };
 
+  const isValidTime = (time: Date) => {
```
benkoppe: This is a useful function but it might be better suited for some sort of shared utility file, especially since it enforces rules that we will probably want to keep in-sync across the codebase.

frontend/src/components/RiderComponents/RequestRideDialog.tsx
```diff
@@ -538,6 +579,21 @@ const RequestRideDialog: React.FC<RequestRideDialogProps> = ({
     );
   };
 
+  useEffect(() => {
+    if (overlayRadius) {
+      setKMLLink(
```
benkoppe: I love how you integrated KML files! But is this the best place to store the constant value of the link? We probably don't want to hard-code it in a `useEffect`.

Additionally, `useEffect` probably isn't needed here at all. Because what you're doing here is effectively making `kmlLink` a derived state of `overlayRadius` -- if it's false, the link is null, if it's true, the link is set to the constant string. A `useMemo` is always better in this case (fully derived state w/ no side effects).

PR #631, opened 2025-12-08
Fix Docker flow to use `pnpm` (+ modernize Docker flow)

benkoppe:
### Summary

This PR intends to bring proper Docker flow to the codebase post-`pnpm`:

- [x] Use `pnpm` in place of `npm` in Dockerfile
- [x] Rework Docker flow:
  - Optimize for small final image size with multi-stage builds
  - Optimize for caching with use of `pnpm fetch` and separate stages for building frontend and backend
- [x] Improve CI for building/pushing Docker in CI:
  - Use `docker/metadata-action` to automatically set tags
  - This allows for large simplification, where the single `ci-container` job can do the role of multiple segmented CI jobs from before. As a result, we can remove one CI job and one composite action (deletion of `cd-workflow.yml` and `docker-build-push.yml`)

### Notes 

Heavy use of multi-stage builds in Docker flow makes comprehension slightly more difficult, but greatly reduces final image size. As a result, this PR will reduce our final Docker image size from ~1.5GB to ~0.5GB.

### DIFF
--- a/frontend/src/components/AddRideButton/AddRideButton.tsx
+++ b/frontend/src/components/AddRideButton/AddRideButton.tsx
@@ -1,9 +1,9 @@
 import React, { useState } from 'react';
-import { Button } from '../FormElements/FormElements';
 import RideDetailsComponent from '../RideDetails/RideDetailsComponent';
 import { createEmptyRide } from '../../util/modelFixtures';
 import { RideType } from '../../types';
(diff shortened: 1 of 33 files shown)


mjaydenkim (approved):
no notes here -- this looks great!! thank you so much for fixing this up :')

PR #636, opened 2025-12-22
Fix N+1 vulnerability, improve speed for batch gets/scans

mjaydenkim:
### Summary

This pull request fixes a massive backend issue with Carriage: previously, loading rides and scanning our database incurred hundreds of API calls. This led to a number of downstream problems:
- Rate limits w/ DynamoDB would get hit very easily
- Loading rides/any databases with a large number of items is very, very slow
- Potential vulnerability (e.g. a malicious actor with access to a backend endpoint could force the system to process a massive number of rides, leading to a DDoS vulnerability; see https://github.com/cornell-dti/cureviews/issues/539 for a description of a similar issue with CUReviews)

Now, instead of the backend looping through every item it intends to scan and populating it, we run a batchGet() operation. This fixes the above problems and dramatically decreases the amount of time it takes to scan for items (e.g. loading the ride schedule after logging into the student page takes ~1 second instead of >20).

### Test Plan
Unfortunately, testing is difficult since the current server tests don't work and fixing them is a next semester issue LOL. I clicked around the app in all three views and nothing seemed to be broken; extra testing as part of reviews would be great.

### DIFF
--- a/server/src/router/common.ts
+++ b/server/src/router/common.ts
@@ -3,35 +3,118 @@ import { ModelType, ObjectType } from 'dynamoose/dist/General';
 import { Condition } from 'dynamoose/dist/Condition';
 import { Item } from 'dynamoose/dist/Item';
 import { Rider } from '../models/rider';
+import { Location } from '../models/location';
+import { Driver } from '../models/driver';
 
-// Helper function to populate riders for ride objects
+// Helper: batch fetch locations, riders, drivers and return maps keyed by id
+async function buildEntityMapsFromSets(
+  locationIds: Set<string>,
+  riderIds: Set<string>,
+  driverIds: Set<string>
+) {
+  const [locationsArr, ridersArr, driversArr] = await Promise.all([
+    locationIds.size
+      ? Location.batchGet(Array.from(locationIds).map((id) => ({ id })))
+      : Promise.resolve([]),
+    riderIds.size
+      ? Rider.batchGet(Array.from(riderIds).map((id) => ({ id })))
+      : Promise.resolve([]),
+    driverIds.size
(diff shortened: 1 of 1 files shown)


server/src/router/common.ts
```diff
@@ -223,13 +306,12 @@ export function deleteById(
   id: string | ObjectType | undefined,
   table: string
 ) {
-  model.get(id || '', (err, data) => {
+  // Use direct delete to avoid an extra GET request (reduces read throughput)
+  model.delete(id || '', (err) => {
```
YottaYocta: Good call

server/src/router/common.ts
```diff
@@ -3,35 +3,118 @@ import { ModelType, ObjectType } from 'dynamoose/dist/General';
...
-// Helper function to populate riders for ride objects
+// Helper: batch fetch locations, riders, drivers and return maps keyed by id
+async function buildEntityMapsFromSets(
+  locationIds: Set<string>,
+  riderIds: Set<string>,
+  driverIds: Set<string>
+) {
+  const [locationsArr, ridersArr, driversArr] = await Promise.all([
+    locationIds.size
+      ? Location.batchGet(Array.from(locationIds).map((id) => ({ id })))
```
YottaYocta: Saw on the DynamoDB documentation that there's a 100 document limit to what you can get with batchget. Do you think we'll run into this constraint? https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html

YottaYocta: With the exception of maybe locations, I doubt there'd be enough riders and drivers to cause issues in testing. Also since we're planning to switch DBs later, this fix works well for our current setup.

PR #638, opened 2025-12-22
Add new unregistered student message

mjaydenkim:
### Summary
CULift/SDS admin requested a new message for unregistered students.

### Test Plan
Other than this, don't think this really needs testing :p
image


YottaYocta (approved):
Looks good!

PR #641, opened 2025-12-26
Fix driver page update status

mjaydenkim:
### Summary 
Previously, drivers couldn't update the status of their rides by clicking on the entry of the rides table despite a modal being there to do so. The modal is now functional and allows non-past rides to be updated to any possible status; the view is updated after edits as well.

### Test Plan 
Simply testing by creating new rides and updating their statuses to ensure that changes are made.
image

### Notes 
Not breaking, but of note: completed rides can now be updated. I think this is helpful: for example, if a misclick occurs and a driver automatically says that a ride is complete but needs to edit it. Past rides still can't be edited.

### DIFF
--- a/frontend/src/components/Modal/modal.module.css
+++ b/frontend/src/components/Modal/modal.module.css
@@ -41,6 +41,12 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
+  /* Keep the title + close button visible when the modal scrolls */
+  position: sticky;
+  top: 0;
+  z-index: 1;
+  padding-bottom: 0.75rem;
+  background-color: white;
 }
 
 .page {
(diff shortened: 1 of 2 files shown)


frontend/src/components/RideDetails/RideActions.tsx
```diff
@@ -98,17 +72,20 @@ const RideActions: React.FC<RideActionsProps> = ({
...
     try {
-      // In a real app, make API call to update status
-      // await updateRideStatus(ride.id, selectedStatus);
-      setUpdateStatusOpen(false);
-      setSelectedStatus(null);
+      await updateRideStatus(ride.id, newStatus);
+      showToast('Ride status updated', ToastStatus.SUCCESS);
+
+      // Ensure any date-based ride views stay in sync with the latest status
+      await refreshRides();
```
YottaYocta: I tried creating a ride and then changing it from the drivers side, but got an error

Image

Image

YottaYocta (changes requested):
Thanks for the PR, and great work updating the UI. I ran into some issues when trying to change the ride, so I think it might be good to take a look.

PR #642, opened 2025-12-26
Remove report button

mjaydenkim:
### Summary 
This PR removes the report button on the student side. This button is nonfunctional in the current version of the app and SSIT suggested that it be removed.

### Test Plan 
None / not necessary.

### Notes 
The idea behind this button, to my knowledge, was reporting issues with drivers or a ride; however, given the small size of CULift in terms of user base/drivers, we think it makes more sense for users to contact admin directly (which there's another button for). Overall, I made this PR because I think the report button isn't necessary, but if anyone disagrees we should discuss in the comments.


YottaYocta (approved):
Yep! I think direct contact is the way to go, especially with all the intricacies involved in scheduling

PR #644, opened 2025-12-28
Fix add student modal

mjaydenkim:
### Summary 

This PR is the first step towards fixing the add student modal.
- [x] Split name into first and last names, removing problems with requiring first and last name with one space separating them
- [x] Removed address, as SSIT said they'd prefer not to immediately include it
- [x] (Made address optional in rider schema -- see below)
- [x] Automatic active/inactive status based on start/end dates
- [x] Added more robust phone number validation (e.g. all regular 10-digit phone number formats now work)

### Test Plan 
Manually tested.
image

### Notes 
As stated, address is now optional; the app previously treated it as optional in the sparing cases where it was used at all so this shouldn't be a breaking issue but I wanted to note it.

### DIFF
--- a/frontend/src/components/Modal/RiderModalInfo.tsx
+++ b/frontend/src/components/Modal/RiderModalInfo.tsx
@@ -169,7 +169,7 @@ const RiderModalInfo: React.FC<ModalFormProps> = ({
     joinDate,
     endDate,
   }) => {
-    const email = netid ? `${netid}@cornell.edu` : undefined;
+    const email = netid ? `${(netid).toLowerCase()}@cornell.edu` : undefined;
     const accessibility = needs.map((option) => option.value.toString());
     const normalizedPhoneNumber = normalizePhoneNumber(phoneNumber);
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/Modal/RiderModalInfo.tsx
```diff
@@ -180,62 +212,87 @@ const RiderModalInfo: React.FC<ModalFormProps> = ({
...
-            type="tel"
+            type="text"
+            disabled={isStudentEditing}
             className={styles.firstRow}
             aria-required="true"
           />
-          {errors.phoneNumber && (
-            <p className={styles.error}>Phone number is not valid</p>
+          {errors.netid && (
+            <p className={styles.error}>NetId cannot be empty</p>
```
YottaYocta: Image
If we have only one error message, what do you think about making it more general? i.e. "Invalid NetID", that way, it covers empty netids and also incorrectly-formatted ones

YottaYocta (changes requested):
Overall, great work! Works as expected on my end. I left some comments related to smaller UI considerations, but other than that this should be good to go

YottaYocta:
Normalization and error for NetIDs look good! Merging

YottaYocta (approved):

PR #647, opened 2026-01-01
Disable map view, fix admin add rides flow

mjaydenkim:
### Summary 
- [x] Implement flag to enable/disable map view (due to custom locations, admins don't want the map view yet)
- [x] Fix admin change location modal (flow was kinda clumsy before)
- [x] Add custom locations to admin add location modal
  - [ ] Note: there isn't any address validation like there is in the student page. This is easy to implement but I was wondering if we should give admins more leverage to input fully custom addresses; worth discussing in reviews though.

### Test Plan 
Tested app functionality with and without flags and scheduling rides from admin side

### DIFF
--- a/frontend/src/components/RideDetails/RideEditContext.tsx
+++ b/frontend/src/components/RideDetails/RideEditContext.tsx
@@ -6,7 +6,7 @@ import React, {
   useEffect,
   ReactNode,
 } from 'react';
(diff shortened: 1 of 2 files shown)


mjaydenkim:
@YottaYocta i think these issues have been fixed, take a look !

PR #650, opened 2026-01-05
Fully Custom Location

anika-4444:
### Summary 

- [x] implemented ability to enter any custom location (does not have to be a real address)
- [x] added custom location placeholder map  

image
image
image

### Notes 
Because custom locations no longer require an address, the Google Maps got kind of wonky. To work around this problem, I added a placeholder for the map that appears when a custom location is used.

Overall, this method could cause more repeat locations than what was there before (which only created new locations based on if the latitude/longitude could not be found in the db), but because we cannot require a real address, I think we will have to just deal with this. One way to mitigate this could be to search the db for locations that have the same name. Will add this asap

### DIFF
--- a/frontend/src/pages/Rider/Schedule.tsx
+++ b/frontend/src/pages/Rider/Schedule.tsx
@@ -153,7 +153,6 @@ const Schedule: React.FC = () => {
   };
 
   const handleRideSubmit = async (formData: FormData) => {
-    console.log('reached handleRideSubmit');
     if (!formData.pickupLocation || !formData.dropoffLocation) return;
 
     // For now, block any recurring rides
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Rider/Schedule.tsx
```diff
@@ -172,6 +179,8 @@ const Schedule: React.FC = () => {
         new Date(startISO).getTime() + 30 * 60 * 1000
       ).toISOString();
 
+      console.log('about to post to rides');
```
mjaydenkim: maybe not necessary?

YottaYocta (approved):
I second @mjaydenkim; these changes look good! @anika-4444 wanted your opinion but should we add a way for the admins to request a custom ride in this PR, or do so in a future one? I noticed that admins could only add rides based on existing locations.

Image

mjaydenkim:
@YottaYocta i believe that’s in my pr #647 which is kinda outdated after this one but i can merge that element in!

PR #683, opened 2026-03-23
Prisma migration

anika-4444:
### Summary 

This is the migration from DyanmoDB to PrismaPostgres. Included in this PR is the schema rewrite and routes changed to use Prisma. There is also a lot of files that Prisma auto-generated during the schema creation. 

When testing, running `npx prisma generate` can be helpful to see the database better formatted. You will also have to create a local instance of the db. 

Recurring rides still needs to be implemented, but I'll do that in a separate PR.

This pull request is the first step towards implementing prisma re-write

- [x] implemented schema change
- [x] changed router files
- [ ] in a separate PR, I'll delete all the dynamo schema leftovers. 

### Breaking Changes 


server/prisma/schema.prisma
```diff
@@ -0,0 +1,250 @@
...
+
+//--------------------
+
+/// A platform administrator who manages riders, drivers, and rides
+model Admin {
+  id          String      @id @default(uuid())
+  firstName   String
+  lastName    String
+  roles       AdminRole[]
+  isDriver    Boolean     @default(false)
```
mjaydenkim: From the admin side, if you have an admin and you check the box to say they're also a driver it returns an error if they're not already a driver. Might be a bit more work but can we merge the employee models (e.g. Admin and Driver)? It would make it a lot easier to handle people who are both; this was a problem in the original schema design, where deleting an employee would delete them from one of the tables at random but not really both.

server/prisma/schema.prisma
```diff
@@ -0,0 +1,250 @@
...
+}
+
+//--------------------
+
+/// A platform administrator who manages riders, drivers, and rides
+model Admin {
+  id          String      @id @default(uuid())
+  firstName   String
+  lastName    String
+  roles       AdminRole[]
```
mjaydenkim: When I try to set employee roles using the dialog it sets and then resets automatically on the UI. Don't know why that is, the request posts correctly, but it's very suspicious.

server/prisma/schema.prisma
mjaydenkim: If I add an employee with my exact details and check the "driver" box it merges the two entries. I think this should be one table.

server/prisma/schema.prisma
```diff
@@ -0,0 +1,250 @@
...
+  active       Boolean     @default(true)
+  joinDate     DateTime    @default(now())
+  rides Ride[]
+}
+
+//--------------------
+
+/// A rider who requests and takes rides
+model Rider {
+  id                String          @id @default(uuid())
```
mjaydenkim, line 136: Good work with these models, they make good sense to me and integrate well with the current app!

server/src/auth/sso-helpers.ts
```diff
@@ -31,74 +29,66 @@ export async function findUserByNetID(
...
-          // IMPORTANT: Check if Rider is active (same as Google OAuth)
-          if (!rider.active) {
-            return { error: 'User not active', userType: 'Rider' };
-          }
-          return { user: rider, userType: 'Rider' };
-        }
-        return null; // User not found in Riders table
-      }
+    if (requestedUserType === 'Rider') {
+      const rider = await prisma.rider.findUnique({
```
mjaydenkim, line 33: I like how much a lot of this server-side code is cleaned up -- great work!

mjaydenkim (changes requested):
Hey -- great work overall! I made a few commits with small changes, this looks great overall though, I did a basic run-through of most app flows and they worked well! Just one major change I'm suggesting (merging the admin/driver tables); other than that I don't see any problems! Good stuff <33

server/src/auth/sso-helpers.ts
mjaydenkim: Forgot to mention this earlier -- any specific plans or outlines on how we plan to host our database?

